PR #111, opened 2017-05-31
feat(react-grid): Implement ColumnOrderState plugin

gsobolev:


packages/dx-react-grid/src/plugins/column-order-state.test.jsx
```diff
@@ -0,0 +1,126 @@
...
+        <ColumnOrderState
+          defaultOrder={['a', 'b']}
+        />
+        <Template
+          name="root"
+          connectGetters={getter => ({ mock: columnsMock(getter('columns')) })}
+          connectActions={action => ({ setColumnOrder: action('setColumnOrder') })}
+        >
+          {({ setColumnOrder }) => (
+            <button
```
dxbykov: Not sure if the button required here. Can't we just save setColumnOrder in a local variable and call it instead of simulating a button click?

PR #123, opened 2017-06-06
refactor(react-grid): Generalize tables

kvet:


packages/dx-react-grid-material-ui/src/templates/table.jsx
```diff
@@ -2,85 +2,66 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import {
-    Table as TableMUI,
-    TableBody as TableBodyMUI,
-    TableHead as TableHeadMUI,
+  Table as TableMUI,
```
dxbykov: Seems like it's unnecessary to introduce MUI aliases now.

kvet: The `Table` component is exported from the file

PR #128, opened 2017-06-09
feat(react-grid): Implement column reordering

kvet:


packages/dx-grid-core/src/utils/table.test.js
```diff
@@ -32,4 +34,38 @@ describe('table utils', () => {
...
+      expect(getTableColumnGeometries(columns, 400)).toEqual([
+        { left: 0, right: 100 },
+        { left: 100, right: 250 },
+        { left: 250, right: 400 },
+      ]);
+    });
+  });
+
+  describe('#getTableTargetColumnIndex', () => {
+    it('should correctly return geometries', () => {
```
gsobolev: Do mean index here?

kvet: Thanks. I'll fix it.

PR #147, opened 2017-06-15
chore: Add missing dependency to the dx-testing package

gsobolev:


packages/dx-testing/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "devDependencies": {
     "eslint": "^3.19.0",
-    "eslint-config-airbnb-base": "^11.1.3"
+    "eslint-config-airbnb-base": "^11.1.3",
+    "eslint-plugin-import": "^2.3.0"
```
dxbykov: This package doesn't use es6 modules syntax. Why is it needed?

gsobolev: This package is a peer dependency of "eslint-config-airbnb-base".
```
$ npm info "eslint-config-airbnb-base@latest" peerDependencies
{ eslint: '^3.19.0', 'eslint-plugin-import': '^2.2.0' }
```

PR #148, opened 2017-06-16
fix(react-grid): Put the dx-react-grid dependencies in order

dxbykov:


packages/dx-react-grid/README.md
```diff
@@ -12,21 +12,27 @@ The DevExtreme React Grid component allows you to display table data with a set
 npm i @devexpress/dx-react-grid --save
 ```
 
-This package does not contain any visual components, so all examples provided below use the DevExtreme React Grid Bootstrap3 package to use Bootstrap rendering for the Grid visual components.
+This package does not contain any visual components; all examples below use the DevExtreme React Grid Bootstrap3 package to use Bootstrap rendering for the Grid visual components.
```
dirk-pieterse: "all examples below _use_ the DevExtreme React Grid Bootstrap3 package to _use_ Bootstrap rendering for the Grid visual components" is a confusing sentence. Consider rewriting it.

dxbykov: We can omit the second part of the sentence since it's obvious that if we use the BS3 package then BS3 will be used for rendering the Grid.

viterobk: all examples below import modules from the DevExtreme React Grid Bootstrap 3 package to use Bootstrap rendering for the Grid visual components

PR #160, opened 2017-06-22
fix(react-grid): Fix incorrect table layout if all columns have fixed width

kvet:


packages/dx-react-grid/src/components/table-layout.jsx
```diff
@@ -91,13 +91,20 @@ export class TableLayout extends React.PureComponent {
       const { columns } = this.props;
       const { sourceColumnIndex, targetColumnIndex } = this.state;
 
-      if (sourceColumnIndex === -1 || targetColumnIndex === -1) return columns;
+      let result = columns;
 
-      const result = columns.slice();
+      const staticWidth = columns.filter(column => column.width === undefined).length === 0;
```
gsobolev: May be it is better to call this const `isStaticWith` or `isFixedWidth` so its name could reflect its value type?

packages/dx-react-grid/src/components/table-layout.jsx
```diff
@@ -91,13 +91,20 @@ export class TableLayout extends React.PureComponent {
...
+      if (staticWidth) {
+        result = result.slice();
+        result.push({ type: 'spring' });
```
gsobolev: Is not it better to use 'flex' instead? And why not to define it as constant?

packages/dx-react-grid/src/components/table-layout.test.jsx
```diff
@@ -229,7 +229,32 @@ describe('TableLayout', () => {
       .toMatchObject({ row: rows[1], column: columns[1], e: {} });
   });
 
-  describe('DnD reordering', () => {
+  describe('spring column', () => {
```
gsobolev: See the comment about the name above.

packages/dx-react-grid/src/components/table-layout.test.jsx
```diff
@@ -229,7 +229,32 @@ describe('TableLayout', () => {
...
+    it('should add spring column if all columns have concrete widths', () => {
```
gsobolev: Concrete?

PR #169, opened 2017-06-28
feat(react-grid): column reordering animation

kvet:


packages/dx-grid-core/src/utils/table.js
```diff
@@ -84,3 +86,61 @@ export const getTableTargetColumnIndex = (columnGeometries, sourceIndex, offset)
...
+          zIndex: 100,
+          position: 'relative',
+        };
+      }
+      return [key, result];
+    })
+    .filter(animation => animation[1].left));
+};
+
+export const filterAnimations = animations => new Map([...animations.entries()]
```
dxbykov: I would call it activeAnimations().

dxbykov: ... or getActiveAnimations()

PR #199, opened 2017-07-13
perf(react-grid): update only changes rows in table

kvet:


packages/dx-react-core/src/index.js
```diff
@@ -5,8 +5,10 @@ export { Getter } from './plugged/getter';
...
 export { Template } from './plugged/template';
 export { TemplatePlaceholder } from './plugged/template-placeholder';
-export { combineTemplates } from './utils/templateHelpers';
 
 export { DragDropContext } from './drag-drop/context';
 export { DragSource } from './drag-drop/source';
 export { DropTarget } from './drag-drop/target';
+
+export { TemplateComponent } from './template-component';
+export { combineTemplates } from './utils/templateHelpers';
```
gsobolev: Why 'templateHelpers' named in camelCase while other files are in kebab-case?

packages/dx-react-core/src/template-component.jsx
```diff
@@ -0,0 +1,2 @@
+export const TemplateComponent = ({ template, children, ...restProps }) =>
```
gsobolev: Let's think of a better name for this entity.

packages/dx-react-core/src/index.js
kvet: It is not related to the current PR.

PR #205, opened 2017-07-19
feat(react-grid): implement grouping by drag and drop

gsobolev:
BREAKING CHANGE:

The TableHeaderRow plugin's `allowGrouping` property has been renamed to `allowGroupingByClick`.


packages/dx-react-demos/src/bootstrap3/featured-virtual-scrolling/demo.jsx
```diff
@@ -84,11 +86,11 @@ export default class Demo extends React.PureComponent {
           }}
         />
 
-        <TableHeaderRow allowSorting allowGrouping />
+        <TableHeaderRow allowSorting allowGrouping allowDragging />
```
kvet: The allowGrouping property may confuse our customers. It is not evident that allowDragging will allow grouping by d'n'd. May be it is a good time to rename it to showGroupingButton?

kvet:
Also, update no data text in the group panel.

PR #218, opened 2017-07-25
fix(react-grid): fix grouping if TableGroupRow is placed before any other table plugin

kvet:


packages/dx-grid-core/src/utils/table.js
```diff
@@ -16,10 +16,19 @@ const getColumnId = column => column.name;
 export const tableColumnKeyGetter = (column, columnIndex) =>
   getTableKeyGetter(getColumnId, column, columnIndex);
 
-export const getTableCellInfo = ({ row, columnIndex, columns }) => {
-  if (row.colspan !== undefined && columnIndex > row.colspan) return { skip: true };
-  if (row.colspan === columnIndex) return { colspan: columns.length - row.colspan };
-  return {};
+export const getTableRowColumnsWithColSpan = (columns, colspan) => {
```
gsobolev: It is not obvious that `colspan` is an index.

PR #227, opened 2017-07-28
refactor(react-grid): change TableRow and TableColumn interfaces structure

kvet:
BREAKING CHANGES:
 
`TableRow` and `TableColumn` interfaces structure has been changed. Now, it wraps original rows and columns of the Grid component instead of patching it.

Before:

```ts
interface TableRow extends Row {
  type?: string;
}
interface TableColumn extends Column {
  type?: string;
}
```

After:

```ts
interface TableRow {
  key: string;
  type: string;
  rowId?: number | string;
  row?: Row;
  height?: number;
}
interface TableColumn {
  key: string;
  type: string;
  column?: Column;
  width?: number;
}
```
 
The `CommandHeadingCellArgs` interface related to the TableEditColumn plugin no longer has `column` and `row` fields.

PR #235, opened 2017-08-02
fix(react-grid): pass style field to editCellTemplate of TableEditRow

### LINKED ISSUE #234: Problem w/ editing when using VirtualTableView
When i click edit on a row  when using VirtualTableView the edit column seems to lose its width and all other columns are pushed to the left of the table, my component construction seems to work fine when using regular TableView though.

kvet:
Fixes #234

### DIFF
--- a/packages/dx-react-grid/src/plugins/table-edit-row.jsx
+++ b/packages/dx-react-grid/src/plugins/table-edit-row.jsx
@@ -40,11 +40,12 @@ export class TableEditRow extends React.PureComponent {
             changeRow: ({ rowId, change }) => action('changeRow')({ rowId, change }),
           })}
         >
-          {({ rowId, row, column, value, changeRow }) =>
+          {({ rowId, row, column, value, changeRow, style }) =>
             editCellTemplate({
               row,
               column,
               value,
+              style,
               onValueChange: newValue => changeRow({
                 rowId,
                 change: {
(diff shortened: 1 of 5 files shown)


packages/dx-react-grid/src/plugins/table-edit-row.jsx
```diff
@@ -40,11 +40,12 @@ export class TableEditRow extends React.PureComponent {
             changeRow: ({ rowId, change }) => action('changeRow')({ rowId, change }),
           })}
         >
-          {({ rowId, row, column, value, changeRow }) =>
+          {({ rowId, row, column, value, changeRow, style }) =>
             editCellTemplate({
               row,
               column,
               value,
+              style,
```
kvet: Tests for this is prepared in #227

packages/dx-react-grid/src/plugins/table-edit-row.jsx
```diff
@@ -68,11 +69,12 @@ export class TableEditRow extends React.PureComponent {
             changeAddedRow: ({ rowId, change }) => action('changeAddedRow')({ rowId, change }),
           })}
         >
-          {({ row, rowId, column, value, changeAddedRow }) =>
+          {({ row, rowId, column, value, changeAddedRow, style }) =>
```
kvet: The same here

PR #237, opened 2017-08-02
feat(react-core): implement dependencies support for PluginContainer

gsobolev:


packages/dx-core/src/utils.js
```diff
@@ -1,13 +1,16 @@
...
+  const aPosition = a.position();
+  const bPosition = b.position();
+  for (let i = 0; i < Math.min(aPosition.length, bPosition.length); i += 1) {
+    if (aPosition[i] < bPosition[i]) return -1;
+    if (aPosition[i] > bPosition[i]) return 1;
+  }
+  return aPosition.length - bPosition.length;
+};
+
+export const insertWithSorting = (newItem, array) => {
```
kvet: Slightly strange API. Typically we specify the array to work with in the first parameter. What do you think?

gsobolev: I was thinking in the way "insert something somewhere". But I can change this in order to stay consistent with other methods.

PR #246, opened 2017-08-07
docs: update the Controlled and Uncontrolled Modes guide

viterobk:


packages/dx-react-grid/docs/guides/controlled-and-uncontrolled-modes.md
```diff
@@ -30,35 +36,36 @@ export class MyApp extends React.PureComponent {
 }
 ```
 
-`Sorting` represents the Grid sorting configuration, and the `changeSorting` function is a handler that is invoked every time the sorting configuration changes. Note that all the state management plugins work with the serializable state. This means that you can persist and restore it in [localStorage](https://developer.mozilla.org/en/docs/Web/API/Window/localStorage) or any other storage that can store string values. The controlled state mode can also be helpful if you need to indicate the current state in your UI or bind controls existing outside the Grid, for example, to put a ComboBox with the available sort orders and let end-users use it for sorting Grid data.
+If you need to apply a new configuration to the Grid, update the parent component state using the `setState` method, which causes the component rendering with new state applied.
```
dirk-pieterse: "which causes the component rendering with new state applied." This part doesn't make sense.

PR #249, opened 2017-08-07
feat(react-grid): adopt PluginContainer dependencies in Grid plugins

gsobolev:


packages/dx-react-grid/src/plugins/local-filtering.jsx
```diff
@@ -8,7 +8,12 @@ export class LocalFiltering extends React.PureComponent {
     const { filterFn } = this.props;
 
     return (
-      <PluginContainer>
+      <PluginContainer
+        pluginName="LocalFiltering"
+        dependencies={[
```
kvet: Typically, creating inlining a complex constant in a render method is not a good practice for performance reasons. Please, consider extracting constants.

packages/dx-react-grid/src/plugins/grouping-panel.jsx
```diff
@@ -13,7 +13,13 @@ export class GroupingPanel extends React.PureComponent {
     } = this.props;
 
     return (
-      <PluginContainer>
+      <PluginContainer
+        pluginName="GroupingPanel"
+        dependencies={[
+          { pluginName: 'GroupingState' },
+          { pluginName: 'SortingState', optional: true },
```
kvet: We already have all information about options that turns dependency to required. Why is it not reflected here?

packages/dx-react-grid/src/plugins/table-header-row.test.jsx
```diff
@@ -26,6 +26,7 @@ const defaultDeps = {
       style: {},
     },
   },
+  plugins: ['EditingState', 'TableView'],
```
SergeyAlexeev: Is `EditingState` correct dependency for `TableHeaderRow`?

PR #261, opened 2017-08-14
fix(react-grid): preserve group panel height when no grouping is specified

gsobolev:


packages/dx-react-grid-material-ui/src/templates/group-panel.jsx
```diff
@@ -12,6 +12,7 @@ const styleSheet = createStyleSheet('GroupPanel', theme => ({
     width: '100%',
   },
   groupInfo: {
+    padding: '6px 0',
```
SergeyAlexeev: Can we use `theme.spacing.unit` here instead of `px`?

PR #325, opened 2017-09-08
refactor(react-grid): switch to TemplateConnector instead of embedded Template connection

kvet:


packages/dx-react-grid/src/plugins/grouping-panel.jsx
```diff
@@ -21,33 +41,22 @@ export class GroupingPanel extends React.PureComponent {
...
+        <Template name="header">
+          <div>
+            <TemplateConnector>
+              {(getters, actions) => (
+                <TemplateRenderer
+                  template={groupPanelTemplate}
+                  params={getGroupPanelTemplateArgs({
+                    getters,
+                    actions,
+                    scope: { allowDragging, allowSorting, allowUngroupingByClick },
```
dxbykov: Let's think if there is a better name for this.

Issue #204, opened 2017-07-18
Parsing error: Unexpected token on simple grid example


DrainMorten:
Hi! This is probably more a setup issue than a Devexpress react grid issue, but maybe its possible to be pointed in a direction.
In our react app I have followed the devexpress react grid documentation, and installed:

npm i --save @devexpress/dx-react-core @devexpress/dx-react-grid
npm i --save @devexpress/dx-react-grid-bootstrap3

and copied in the simple grid starting code, but I get a parsing error: Unexpected Token on the third '}' in the rows line: rows={[{ id: 0, ... }, ...]}

Sorry if this is very vague, it just does not say much about this type of setup issues in the the documentation. The code is basically the same as the example, but posted a screenshot of the parsing error for easier viewing.

![dxgrid](https://user-images.githubusercontent.com/29375412/28320662-78d77442-6bd1-11e7-8378-2f6ab9e940f7.png)

Any tips is appreciated!
Morten

kvet:
Hi,

`...` means that there may be any number of other fields or array items. You can check our [WebpackBin for Bootstrap3](https://www.webpackbin.com/bins/-KoCcFwPPtps5fEN0PkI) without the `...` symbol.

We will update our basic with less confusing code.

DrainMorten:
Ah, right, thanks! Still learning..

Another issue I encountered when installing bootstrap 3 package from the documentation, only with yarn:
yarn add --dev @devexpress/dx-react-grid-bootstrap3

It seemed to install ok but I got the message:
warning "@devexpress/dx-react-grid-bootstrap3@1.0.0-alpha.5" has unmet peer dependency "react-bootstrap@^0.31.0".

Then when trying to compile i get this error:
ERROR in ./~/@devexpress/dx-react-grid-bootstrap3/dist/dx-react-grid-bootstrap3.es.js
Module not found: Error: Can't resolve 'react-bootstrap' in '...(local file path)...\node_modules\@devexpress\dx-react-grid-bootstrap3\dist'

Is there usually some other places in an application this must be setup also?

DrainMorten:
SOLVED: I solved the above issue by manually installing the react-bootstrap@^0.31.0.

kvet:
I've noticed that we don't require installing of the 'react-bootstrap' package in the documentation. But it is required as DevExtreme React Grid uses it in the Bootstrap 3 templates package. 

Thank you for reporting this issue. We will fix it.

PR #336, opened 2017-09-14
feat(react-grid): support column data types

gsobolev:


packages/dx-react-grid-bootstrap3/src/templates/table-filter-cell.jsx
```diff
@@ -1,30 +1,34 @@
...
+      />
+    )}
   </th>
 );
 
 TableFilterCell.propTypes = {
   style: PropTypes.object,
   filter: PropTypes.object,
   setFilter: PropTypes.func,
+  children: PropTypes.node,
```
kvet: What about node list?

packages/dx-react-grid/docs/guides/data-types.md
```diff
@@ -0,0 +1,69 @@
...
+- [DataTypeProvider](../reference/data-type-provider.md) - the plugin which provides abilities for customization
+
+## Formatting
+
+To format values in cells of a particular column add the [DataTypeProvider](../reference/data-type-provider.md) and set its `type` according to the `dataType` of the column. Then pass the formatting function to the `formatterTemplate` property.
+
+```js
+const rows = [
+  { product: 'SolarOne', price: '3039' },
+  ...
```
kvet: Avoid `...`. See: #204

packages/dx-react-grid/src/plugins/data-type-provider.jsx
```diff
@@ -0,0 +1,38 @@
...
+import PropTypes from 'prop-types';
+import { PluginContainer, Template } from '@devexpress/dx-react-core';
+
+// eslint-disable-next-line react/prefer-stateless-function
+export class DataTypeProvider extends React.PureComponent {
+  render() {
+    const { formatterTemplate, editorTemplate, type } = this.props;
+    return (
+      <PluginContainer name="DataTypeProvider">
+        <Template
```
kvet: Should we skip core template registration if the user template is not defined?

packages/dx-react-grid/src/plugins/table-filter-row.jsx
```diff
@@ -17,10 +21,18 @@ const getFilterTableCellTemplateArgs = (
 ) => ({
   ...params,
   column: params.tableColumn.column,
+  row: params.tableRow.row,
```
kvet: Why do you pass a row here? It is null for the TableFilterRow and it also is not documented.

packages/dx-react-grid/src/plugins/table-group-row.jsx
```diff
@@ -29,6 +33,12 @@ const getGroupTableCellTemplateArgs = (
   toggleGroupExpanded: () => toggleGroupExpanded({ groupKey: params.tableRow.row.key }),
 });
 
+const getValueEditorArgs = params => ({
```
kvet: Why editor?

packages/dx-react-grid/src/plugins/data-type-provider.test.jsx
```diff
@@ -0,0 +1,81 @@
...
+  //   expect(valueFormatterPredicate({ column: { dataType: 'value' } }))
+  //     .toBeFalsy();
+
+  //   const valueEditorPredicate = tree.findWhere(n => n.prop('name') === 'valueEditor')
+  //     .prop('predicate');
+  //   expect(valueEditorPredicate({ column: { dataType: 'test' } }))
+  //     .toBeTruthy();
+  //   expect(valueEditorPredicate({ column: { dataType: 'value' } }))
+  //     .toBeFalsy();
+  // });
```
kvet: ???

packages/dx-react-grid/src/plugins/table-filter-row.jsx
```diff
@@ -21,7 +21,6 @@ const getFilterTableCellTemplateArgs = (
 ) => ({
   ...params,
   column: params.tableColumn.column,
-  row: params.tableRow.row,
```
dsibiski: @gsobolev I'm thinking there must be some specific reason (that I don't know) as to why you removed this. Would you please elaborate?

I'm asking because in my use case, it would be extremely useful to have the rest of the row, or really the ID of the row so that I can access another field in my data. I'm not keep on having formatting code in my react component but instead I'd like to pre-format all of my data before it gets to the grid. In this case, my "formatter" would simply choose another field from the data to display.

Any thoughts would be very useful! Thanks for this amazing library!

gsobolev: Hi @dsibiski!
 
I am not sure I understand your concerns. The `getFilterTableCellTemplateArgs` is a helper which returns parameters only for the filter row cells. The filter row itself does not hold any data fields.
Please provide more details on your use case so I can understand your scenario.
Let's also continue this discussion as a separate issue rather than in this pull request.

dsibiski: @gsobolev Thanks for the reply, sorry for confusing the code (thought I was looking at something else). I'll open up a proper issue and explain my use case better there.

PR #345, opened 2017-09-19
feat(react-demos): add custom local filtering demo

gsobolev:


packages/dx-react-grid/docs/guides/filtering.md
```diff
@@ -9,35 +9,39 @@ The following plugins implement filtering features:
 - [LocalFiltering](../reference/local-filtering.md) - performs local data filtering
 - [TableFilterRow](../reference/table-filter-row.md) - renders a filter row
 
-Note that [plugin order](../README.md#plugin-order) is important.
+Note that [plugin order](./plugin-overview.md#plugin-order) is important.
```
kvet: Hm, it seems is broken some time ago across all guides... Can you prepare a separate PR with a fix?

gsobolev: Ok.

packages/dx-react-grid/docs/guides/filtering.md
```diff
@@ -11,33 +11,37 @@ The following plugins implement filtering features:
...
 .embedded-demo(filtering/local-filtering-controlled)
 
-## Filter Row Customization
+### Using Custom Filtering Algorithms
 
-Define a filter row cell template using the `TableFilterRow` plugin's `filterCellTemplate` property to use a custom editor instead of the built-in one. Pass a function that renders a custom component to the `filterCellTemplate` property. Define the component's state and handle its changes using the `filter` and `setFilter` arguments respectively to delegate the component's state control to the `TableFilterRow` plugin. In this case, the `FilteringState` plugin handles changes internally, which allows you to use the Grid in both controlled and uncontrolled modes.
+You can also specify a filtering predicate using the `LocalFiltering` plugin's `filterFn` property to implement a custom filtering logic.
 
-## Custom Filtering Algorithm
+Alternatively, it is possible to go along without the `LocalFiltering` plugin. In this case, you should filter rows manually on the `FilteringState` plugin's `onFiltersChange` event and then pass the result to the Grid component's `row` property.
```
dxbykov: It's not quite clear from the description when I should use filterFn and when I should skip using the LocalFiltering plugin.

dxbykov: I think it's not also properly to say that you should filter your rows on the 'onFiltersChange' event. It might appear that the rows change on some other piece of state changes. In this case, you should also filter the rows. Actually, it's shown in your demo. You filter rows in the render method (once state or props have changed).

PR #350, opened 2017-09-20
docs(react-grid): update filtering guide

gsobolev:


packages/dx-react-grid/docs/guides/filtering.md
```diff
@@ -11,33 +11,29 @@ The following plugins implement filtering features:
```
dirk-pieterse: "The filtering state management, Filter Row rendering, and filtering logic are implemented in the following plugins:" and then
"The following plugins implement filtering features:"? Why do you repeat this?

gsobolev: Looks like there is no need in the last sentence of the first paragraph. I'll remove it. Thanks!

packages/dx-react-demos/src/material-ui/filtering/custom-filter-row.jsx
```diff
@@ -16,81 +13,73 @@ import {
...
-  filter: PropTypes.object,
 };
 
-SexFilterCellBase.defaultProps = {
+UnitsFilterCellBase.defaultProps = {
   filter: null,
 };
 
-const SexFilterCell = withStyles(styles, { name: 'SexFilterCell' })(SexFilterCellBase);
+const UnitsFilterCell = withStyles(styles, { name: 'SexFilterCell' })(UnitsFilterCellBase);
```
YashZolmajdi: missed `{ name: 'SexFilterCell' }`, Looks so out of place in official [demo documentation](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/filtering/)

SergeyAlexeev: @YashZolmajdi thank you for pointing that. The related PR #1750

PR #354, opened 2017-09-21
refactor(react-grid): change custom filtering function signature

kvet:
BREAKING CHANGE:
The 'filterFn' property of the LocalFiltering has been renamed to 'predicate'. The argument list has been changed from `filterFn(row: Row, filter: Filter) => boolean` to  `predicate(value: any, filter: Filter, row: Row) => boolean`.

### DIFF
--- a/packages/dx-grid-core/src/plugins/filtering-state/computeds.js
+++ b/packages/dx-grid-core/src/plugins/filtering-state/computeds.js
@@ -1,13 +1,15 @@
-const toString = value => String(value).toLowerCase();
+const toLowerCase = value => String(value).toLowerCase();
 
-const applyFilter = (filter, value) => (toString(value).indexOf(toString(filter.value)) > -1);
-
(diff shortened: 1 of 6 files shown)


packages/dx-grid-core/src/plugins/filtering-state/computeds.js
```diff
@@ -1,13 +1,15 @@
...
-
-const applyFilter = (filter, value) => (toString(value).indexOf(toString(filter.value)) > -1);
-
-export const filteredRows = (rows, filters, getCellData, userFilterFn) => {
+const toLowerCase = value => String(value).toLowerCase();
+
+export const filteredRows = (
+  rows,
+  filters,
+  getCellData,
```
SergeyAlexeev: getCellValue?

kvet: Another PR...

PR #355, opened 2017-09-21
perf(react-grid): suppress redundant render

### LINKED ISSUE #353: Perfomance issue when editing
I'm using code example from material-ui/editing/edit-row-controlled.jsx with my custom data (20 rows, 7 cols).  In row edit mode when I type a symbol it takes ~0.5 sec to be shown in input. When I check "Highlight Updates" in react developer tools in chrome, I see that with every typed symbols the whole grid is updated.

What can be the cause? Thanks in advance.

kvet:
related to #353


packages/dx-react-grid-bootstrap3/src/templates/table-edit-command-cell.jsx
```diff
@@ -42,12 +42,13 @@ export const EditCommandHeadingCell = ({
 EditCommandHeadingCell.propTypes = {
   addRow: PropTypes.func.isRequired,
   commandTemplate: PropTypes.func.isRequired,
-  addCommandText: PropTypes.string.isRequired,
+  addCommandText: PropTypes.string,
   allowAdding: PropTypes.bool.isRequired,
   style: PropTypes.object,
 };
```
gsobolev: Please add an empty line here like you do in other files.

packages/dx-react-grid-bootstrap3/src/templates/table-edit-command-cell.jsx
```diff
@@ -117,12 +118,16 @@ EditCommandCell.propTypes = {
...
-  editCommandText: PropTypes.string.isRequired,
-  deleteCommandText: PropTypes.string.isRequired,
-  commitCommandText: PropTypes.string.isRequired,
-  cancelCommandText: PropTypes.string.isRequired,
+  editCommandText: PropTypes.string,
+  deleteCommandText: PropTypes.string,
+  commitCommandText: PropTypes.string,
+  cancelCommandText: PropTypes.string,
   style: PropTypes.object,
 };
```
gsobolev: And here.

PR #356, opened 2017-09-22
perf(react-grid): optimize internal state calculation

kvet:
related to #353


packages/dx-react-demos/src/bootstrap3/editing/edit-row-controlled.jsx
```diff
@@ -36,6 +36,7 @@ export default class Demo extends React.PureComponent {
       changedRows: {},
     };
 
+    this.getRowId = row => row.id;
```
dxbykov: It's pure function. Why do we need to recreate it for each component instance? I would declare it as a const outside the component (looks like we did it this way in other demos).

PR #367, opened 2017-09-27
docs(react-grid): update DataTypeProvider related docs

gsobolev:


packages/dx-react-grid/docs/reference/data-type-provider.md
```diff
@@ -47,7 +47,7 @@ A value with the following shape:
 Field | Type | Description
 ------|------|------------
 column | [Column](#column) | A column object.
-row | [Row](grid.md#row) | A row object.
+row? | [Row](grid.md#row) | A row object.
```
kvet: What about the same with a formatter?

PR #394, opened 2017-10-11
docs: update the GroupingPanel plugin reference

viterobk:


packages/dx-react-grid/docs/reference/grouping-panel.md
```diff
@@ -26,7 +26,7 @@ groupPanelItemTemplate? | (args: [GroupPanelItemProps](#group-panel-item-props))
 
 ### <a name="column"></a>Column (Extension)
 
-A value with the [Column](grid.md#column) shape extended by the following fields:
+A value with the [Column](grid.md#column) shape extended by the following field:
```
kvet: Can we revert this change? We can add an additional field in the future and miss to update this line.

PR #395, opened 2017-10-11
feat(react-grid): support custom grouping processing

kvet:


packages/dx-grid-core/src/plugins/custom-grouping/computeds.test.js
```diff
@@ -0,0 +1,247 @@
...
+        ],
+      }, {
+        key: 2,
+        items: [
+          { a: 2, b: 1 },
+          { a: 2, b: 2 },
+        ],
+      }];
+      const getHierarchicalChildGroups = groups => groups
+        .map(group => ({ key: String(group.key), value: group.key, nestedData: group.items }));
```
SergeyAlexeev: Can we extract duplicated code? I mean something like this:
```js
const createGetChildGroupsMock = (groups) => {
  const getHierarchicalChildGroups = groups => groups
        .map(group => ({ key: String(group.key), value: group.key, nestedData: group.items }));
  return jest.fn(getHierarchicalChildGroups);
};
```

kvet: I don't like decreasing tests readability as it will not reduce loc count dramatically.

packages/dx-react-demos/src/bootstrap3/grouping/local-grouping-static.test.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import Demo from './local-grouping-static';
+
+describe('BS3: grouping with custom', () => {
```
SergeyAlexeev: The test name sounds strange

packages/dx-react-demos/src/material-ui/grouping/custom-grouping-static.jsx
```diff
@@ -0,0 +1,68 @@
...
+        { name: 'name', title: 'Name' },
+        { name: 'sex', title: 'Sex' },
+        { name: 'city', title: 'City' },
+        { name: 'car', title: 'Car' },
+      ],
+      grouping: [{ columnName: 'sex' }],
+      data: [{
+        key: 'Male',
+        items: generateRows({
+          columnValues: { ...defaultColumnValues, sex: ['Female'] },
```
SergeyAlexeev: Is this `sex: ['Female']` correct for `Male` group?

PR #439, opened 2017-10-24
feat(react-grid): support expanding/collapsing group rows with keyboard

gsobolev:


packages/dx-react-grid-bootstrap3/src/templates/table-group-row-cell.jsx
```diff
@@ -3,27 +3,35 @@ import PropTypes from 'prop-types';
...
-      cursor: 'pointer',
-      ...style,
-    }}
-    onClick={toggleGroupExpanded}
-  >
-    <i
-      className={`glyphicon glyphicon-triangle-${isExpanded ? 'bottom' : 'right'}`}
+}) => {
+  const handleKeyDown = (event) => {
+    if (event.keyCode === 13) toggleGroupExpanded();
```
kvet: Extract const with readable name, please.

kvet: We don't need to prevent default behaviour or stop event propagation?

packages/dx-react-grid-material-ui/src/templates/table-group-row-cell.jsx
```diff
@@ -1,15 +1,12 @@
...
 import ExpandMore from 'material-ui-icons/ExpandMore';
-
-import {
-  TableCell,
-} from 'material-ui';
-
+import { TableCell } from 'material-ui/Table';
 import { withStyles } from 'material-ui/styles';
 
+const ENTER_KEY_CODE = 13;
```
SergeyAlexeev: Can we extract key code const(s) to a separate file and share it between bootstrap & MUI?

PR #469, opened 2017-11-09
chore(react-grid-material-ui): allow expand rows by group row click

MaximKudriavtsev:


SergeyAlexeev:
Let's move the `onClick` event handler to TableCell and keep `cursor: pointer`.

PR #476, opened 2017-11-10
chore: add issue template

kvet:


artem-kurchenko:
>> If you have DevExtreme Commercial license, you may write to support center (https://www.devexpress.com/ask) in order to receive an answer shortly.

If you own an active DevExtreme license, feel free to contact us in the Support Center (https://www.devexpress.com/ask) to receive an answer shortly.

I'd prefer "active", because previous versions don't have React, Also I belive we don't need to mention "Commercial" here.

PR #486, opened 2017-11-15
refactor(react-grid): replace render functions with components in Grid

kvet:
BREAKING CHANGES:
 
The Grid's `rootTemplate`, `headerPlaceholderTemplate`, and `footerPlaceholderTemplate` properties have been replaced with `rootComponent`, `headerPlaceholderComponent`, and `footerPlaceholderComponent`. This also means that they accept components instead of render functions. Find more details here: https://github.com/DevExpress/devextreme-reactive/issues/496
 
The `headerTemplate`, `bodyTemplate`, and `footerTemplate` properties have been replaced with the `children` property in `rootTemplate`.


packages/dx-react-grid/src/grid.jsx
```diff
@@ -32,19 +32,19 @@ Grid.propTypes = {
   getRowId: PropTypes.func,
   getCellValue: PropTypes.func,
   columns: PropTypes.array.isRequired,
-  rootTemplate: PropTypes.func.isRequired,
-  headerPlaceholderTemplate: PropTypes.func,
-  footerPlaceholderTemplate: PropTypes.func,
+  rootComponent: PropTypes.func.isRequired,
```
gsobolev: Is `PropTypes.func` correct for this property?

kvet: Yes. This is the way React components can be validated

packages/dx-react-grid/docs/reference/grid.md
```diff
@@ -29,25 +29,23 @@ Field | Type | Description
 name | string | Specifies the column name or the name of a row field whose value the column displays. If the column name does not match any field name, specify the `getCellValue` function.
 getCellValue | (row: any, columnName: string) => any | Specifies the function used to get the column value for a given row.
 
-### <a name="root-args"></a>RootArgs
+### RootProps
 
 Describes properties passed to the root template when it is being rendered.
```
SergeyAlexeev: 'passed to the root template' -> 'passed to the root component' ?

PR #494, opened 2017-11-20
fix(react-grid): prevent click event propagation on detail toggle

### LINKED ISSUE #492: [Bug] Stop propagation on details button click.
I have used [such](https://github.com/DevExpress/devextreme-reactive/blob/v1.0.0-alpha.14/packages/dx-react-demos/src/bootstrap3/basic/table-row-template.jsx) way pass an event on body row click, but lately, when you moved this button to material-ui `IconButton` it started to propagate event on click. 
cropped-stop-event-bug

## Expected Behavior
It should not dispatch event on the BodyRow

## Steps to reproduce
1. Create custom row template, and add onClick event to the `TableRow`
2. Add detail plugin to the data grid
3. Click on the expand detail view button
4. Refer that onClick event was dispatched.

It should be easily fixed on adding `e.stopPropagation` to Icon button

## Environment 
| Tech         | Version |
|--------------|---------|
| Material-UI  |    ^1.0.0-beta.21     |
| React        |   ^16.0.0    |
| dx-react-core      |   ^1.0.0-beta.1      |
| dx-react-grid        | ^1.0.0-beta.1        |
| dx-react-grid-material-ui | ^1.0.0-beta.1        |

MaximKudriavtsev:
fixes #492

### DIFF
--- a/packages/dx-react-grid-bootstrap3/src/templates/table-detail-toggle-cell.test.jsx
+++ b/packages/dx-react-grid-bootstrap3/src/templates/table-detail-toggle-cell.test.jsx
@@ -17,9 +17,9 @@ describe('TableDetailToggleCell', () => {
       />
     ));
 
-    const clickHandlerButton = tree.find('td').prop('onClick');
+    const buttonClickHandler = tree.find('td').prop('onClick');
 
-    clickHandlerButton(mockEvent);
+    buttonClickHandler(mockEvent);
     expect(toggleExpanded)
       .toHaveBeenCalled();
     expect(mockEvent.stopPropagation)
--- a/packages/dx-react-grid-bootstrap3/src/templates/table-detail-toggle-cell.test.jsx
+++ b/packages/dx-react-grid-bootstrap3/src/templates/table-detail-toggle-cell.test.jsx
@@ -1,13 +1,33 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { mount, shallow } from 'enzyme';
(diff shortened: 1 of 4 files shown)


packages/dx-react-grid-bootstrap3/src/templates/table-detail-toggle-cell.test.jsx
```diff
@@ -1,13 +1,33 @@
...
+    const mockEvent = {
+      stopPropagation: jest.fn(),
+    };
+    const tree = shallow((
+      <TableDetailToggleCell
+        toggleExpanded={toggleExpanded}
+      />
+    ));
+
+    const clickHandlerButton = tree.find('td').prop('onClick');
```
SergeyAlexeev: `clickHandlerButton` looks like an element, but in fact, it is a function. So, it can be renamed to `buttonClickHandler`.

packages/dx-react-grid-material-ui/src/templates/table-detail-toggle-cell.test.jsx
```diff
@@ -1,39 +1,46 @@
...
+      stopPropagation: jest.fn(),
+    };
+    const tree = shallow((
       <TableDetailToggleCell
         toggleExpanded={toggleExpanded}
       />
     ));
 
-    tree.find('IconButton').simulate('click');
+    const clickHandlerButton = tree.find(IconButton).prop('onClick');
```
SergeyAlexeev: The same rename can be applied here.

PR #510, opened 2017-11-27
Correct grid reference url

paiboon15721:
if add .md extension, in the document will get 404 page not found.

### DIFF
--- a/packages/dx-react-grid/docs/guides/data-types.md
+++ b/packages/dx-react-grid/docs/guides/data-types.md
@@ -20,7 +20,7 @@ const rows = [
 ];
 const columns = [
   { name: 'product', title: 'Product' },
-  { name: 'price', title: 'Sale Price', dataType: 'currency' },
+  { name: 'amount', title: 'Sale Amount', dataType: 'currency' },
 ];
 <Grid
   rows={rows}
--- a/packages/dx-react-grid/docs/guides/data-types.md
+++ b/packages/dx-react-grid/docs/guides/data-types.md
@@ -58,7 +58,7 @@ const columns = [
         value={value}
         onChange={e => onValueChange(e.target.value === 'true')}
       >
-        <option value={false}>No</options>
+        <option value={false}>No</option>
         <option value>Yes</option>
       </select>
     )}
--- a/packages/dx-react-grid/docs/guides/data-types.md
+++ b/packages/dx-react-grid/docs/guides/data-types.md
@@ -20,7 +20,7 @@ const rows = [
 ];
 const columns = [
   { name: 'product', title: 'Product' },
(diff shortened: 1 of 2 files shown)


packages/dx-react-grid/docs/guides/plugin-overview.md
```diff
@@ -9,7 +9,7 @@ The plugins that provide all the Grid's functionality can be divided into four g
 
 Note that the plugins are composable and can be nested into each other.
 
-Refer to the [Reference](../../reference/grid.md) to see the complete plugin list.
+Refer to the [Reference](../../reference/grid) to see the complete plugin list.
```
kvet: Relative '.md' references should point to '.md' files. Our site build system transforms '.md' links to normal links. In order to solve this issue, please change the link path to: '../reference/grid.md'.

SergeyAlexeev: `*.md ` is OK. The problem is the 'reference' folder path is incorrect here. `../../reference/` should be replaced with `../reference/`. Well, the correct link is `../reference/grid.md`

packages/dx-react-grid/docs/guides/data-types.md
```diff
@@ -20,7 +20,7 @@ const rows = [
 ];
 const columns = [
   { name: 'product', title: 'Product' },
-  { name: 'amount', title: 'Sale Amount', dataType: 'currency' },
+  { name: 'price', title: 'Sale Price', dataType: 'currency' },
```
SergeyAlexeev: Is it really valuable change?

kvet: Not sure that this change should be merged into master. It will be better to revert this line.

SergeyAlexeev: Let's keep this piece of code consistent with the [demo](https://github.com/DevExpress/devextreme-reactive/blob/v1.0.0-beta.1/packages/dx-react-demos/src/material-ui/data-types/formatters.jsx#L42).

I think it would be better if we rename `price` to `amount` in the following code sample:

```js
const rows = [
  { product: 'SolarOne', price: '3039' },
];
```

packages/dx-react-grid/docs/guides/plugin-overview.md
paiboon15721: reverted :)

packages/dx-react-grid/docs/guides/data-types.md
paiboon15721: but the row price property doesn't match column name. Sale Amount or price can't display on the grid.

capture

kvet: Oh. I see. My fault. Let's rename the `price` field to `amount` in the `rows` constant.

Issue #496, opened 2017-11-21
Replace render functions with components


kvet:
Starting with the initial release, our components and plugins contain "...Template" properties, which hold render functions. During alpha releases, we added many features providing rich customization capabilities. Now, many of those render functions are treated as React components: they can have children and an associated key property.
 
The Grid wraps each render function in the TemplateRenderer component to support passing an associated key property to this function. This causes creation of multiple TemplateRenderer components in the virtual DOM tree, which decreases performance. We are going to fix this problem. However, it will lead to a breaking change.
 
# Suggested Change
 
- use `...Component` properties that accept a component instead of `...Template` properties that accept render functions
- support component overriding for a specific column for some properties (for example, tableCellComponent in TableView) 
 
# Difference between the Component and Render Function
 
Here is a property receiving a component:
 
```js
<DataTypeProvider
  type="uppercase"
  valueFormatterComponent={ValueFormatter}
/>

const ValueFormatter = ({ value }) => {
  return <span>{value.toUpperCase()}</span>
}
```
 
and here is a render function:
 
```js
<DataTypeProvider
  type="uppercase"
  // NOTE: it is possible to inline child component render
  valueFormatterRender={({ value }) => {
    return <span>{value}</span>
  }}
/>
```
 
If you use a property receiving a component, DataTypeProviter uses React.createElement to create a new React element based on the provided component. It means that if you pass a function to the component attribute, a new component is created each time the Grid is rendered. This causes redundant mounting and unmounting of the component instead of just updating it.
 
# Passing a Render Function as a Component
 
We don't recommend passing a rendering function to a component property because of the issue described above. However, you can work around this limitation using any state management library. For example, Redux.
 
If you want to pass the rendering function to a component property and do not want to use a state management library, we advise you to use the following solution to avoid component unmounting.
 
The original code:
 
```js
// NOTE: we need rows property in the following component,
// but it is not passed by React Grid
const TableRow = ({ children, row, rows }) => (
  <tr
    onClick={() => alert(rows.indexOf(row))}
  >
    {children}
  </tr>
);

export default class Demo extends React.PureComponent {
  constructor(props) {
    super(props);

    this.state = {
      columns: [/* ... */],
      rows: [/* ... */],
    };
  }
  render() {
    const { rows, columns } = this.state;

    return (
      <Grid
        rows={rows}
        columns={columns}
      >
        <TableView
          tableRowComponent={TableRow}
        />
        <TableHeaderRow />
[...]

PR #514, opened 2017-11-27
refactor(react-grid): replace render functions with components in ColumnChooser

gsobolev:
BREAKING CHANGE

The ColumnChooser plugin's  `containerTemplate ` and `itemTemplate ` properties have been replaced with `containerComponent ` and `itemComponent ` ones, which accept components instead of render functions. Find more details here: #496


packages/dx-react-grid/docs/reference/column-chooser.md
```diff
@@ -11,34 +11,34 @@ Name | Type | Default | Description
...
 
 Describes properties passed to the container template when it is being rendered.
 
 A value with the following shape:
 
 Field | Type | Description
 ------|------|------------
-items | Array&lt;[ColumnChooserItem](#column-chooser-item)&gt; | An array of column chooser items.
-onItemToggle | (item: [ColumnChooserItem](#column-chooser-item)) => void | Handles item visibility changes.
+items | Array&lt;[ColumnChooserItem](#columnchooseritem)&gt; | An array of column chooser items.
```
kvet: Hm, where is the children property?

PR #527, opened 2017-11-29
refactor(react-grid): replace render functions with components in PagingPanel

gsobolev:
BREAKING CHANGE

The PagingPanel `pagerTemplate` property has been replaced with `pagerComponent` one, which accepts a component instead of a render function. Find more details here: #496


packages/dx-react-grid/src/plugins/paging-panel.test.jsx
```diff
@@ -44,21 +46,19 @@ describe('PagingPanel', () => {
     jest.resetAllMocks();
   });
 
-  it('should render the "pagerTemplate" template in the "footer" template placeholder', () => {
-    const pagerTemplate = jest.fn().mockReturnValue(null);
-    const deps = {};
-
-    mount((
+  it('should render the "pagerComponent" template in the "footer" template placeholder', () => {
```
SergeyAlexeev: "pagerComponent" template -> "pagerComponent"

PR #541, opened 2017-11-30
fix(react-core): fix drag by mouse on a touch-enabled devices

### LINKED ISSUE #535: Drag&Drop with Laptops which has Touch support
During our trial of this grid we discovered that users which have laptops with touch support could not drag&drop in Chrome. For instance grouping, rearranging columns and resizing columns. When these users are testing in Edge it works fine.

- [ x] I have searched this repository's [issues](https://github.com/devexpress/devextreme-reactive/issues) and believe that this is not a duplicate.

### Current Behaviour
Nothing happens when dragging columns on touch enabled laptops and Chrome (might be more)

### Expected Behaviour
Should behave the same way as normal 

### Your Environment

devextreme-reactive: 1.0.0-beta.1
react: ^16.0.0
browser: Chrome/Edge/++
bootstrap: none
react-bootstrap: none
material-ui: 1.0.0-beta.21

kvet:
fixes #535


packages/dx-react-core/src/draggable.test.jsx
```diff
@@ -178,6 +178,27 @@ describe('Draggable', () => {
         .toHaveBeenCalledWith({ x: 10, y: 10 });
     });
 
+    it('should not fire the "onStart" callback on before timeout in conjunction with mouse event', () => {
```
gsobolev: Please make this description more clear

PR #544, opened 2017-12-01
fix(react-core): remove redundant markup elements

kvet:
BREAKING CHANGES:

To get rid of redundant 'div' elements in markup, we have updated the minimum React version the grid requires to 16.2.0.


packages/dx-react-core/src/plugged/plugin-indexer.jsx
```diff
@@ -8,7 +8,7 @@ export const PluginIndexer = (
   { children },
   { positionContext },
 ) => (
-  <div style={{ display: 'none' }}>
+  <React.Fragment>
```
SergeyAlexeev: What about `display: none`? Is it unnecessary here?

kvet: Fragment is an virtual element. It will not be present in DOM.

PR #545, opened 2017-12-04
refactor(react-core): replace Watcher with the Getter

MaximKudriavtsev:
BREAKING CHANGE:

To simplify working with plugins, we have got rid of the Watcher. 
- Now, you can use a Getter instead.

Before:
```
...
<Watcher
  watch={
    getter = getter('someGetter')
  }
  onChange={(action, someGetter) => {
    action('someAction')(someGetter);
  }}
/>
...
```

After:
```
...
<Getter
  name="someGetter"
  computed={(getters, actions) => {
      actions.someAction(getters.someGetter);
    return getters.someGetter;
  }}
/>
...
```


packages/dx-react-core/src/plugged/template-connector.jsx
```diff
@@ -19,30 +19,38 @@ export class TemplateConnector extends React.Component {
...
-
-    const { getters, trackedDependencies } = getAvaliableGetters(pluginHost);
-    this.trackedDependencies = trackedDependencies;
-    const actions = getAvaliableActions(pluginHost);
-
-    return children(getters, actions);
+    return this.state.children;
   }
 }
-
```
kvet: We leave a blank line between component and property types.

packages/dx-react-grid/src/plugins/local-paging.jsx
```diff
@@ -11,6 +11,13 @@ const rowsWithHeadersComputed = ({ rows, pageSize, getRowLevelKey }) =>
 const totalCountComputed = ({ rows }) => rowCount(rows);
 const paginatedRowsComputed = ({ rows, pageSize, currentPage }) =>
   paginatedRows(rows, pageSize, currentPage);
+const gettersChangesComputed = (returnedValue, getters, actions) => {
```
kvet: Why do you use this several times? Seems like it should be applied only to one getter. Rework it.

PR #555, opened 2017-12-06
refactor(react-grid): remove totalCount property from PagingState to CustomPaging

MaximKudriavtsev:
BREAKING CHANGES:

We have moved the PagingState plugin's `totalCount` property to the CustomPaging plugin to improve API consistency. Now, you can add the CustomPaging plugin to a project and specify its `totalCount` property to implement remote paging.

Before:

```js
<Grid>
  <...>
  <PagingState totalCount={10} />
</Grid>
```

After:

```js
<Grid>
  <...>
  <PagingState />
  <CustomPaging totalCount={10} />
</Grid>
```


packages/dx-react-grid/src/plugins/paging-state.jsx
```diff
@@ -75,5 +72,5 @@ PagingState.defaultProps = {
   currentPage: undefined,
   defaultCurrentPage: 0,
   onCurrentPageChange: undefined,
-  totalCount: 0,
+  // totalCount: 0,
```
kvet: Remove this

ogheneovo12:
Custom Paging causes a bug with page sizing and rows displayed

PR #557, opened 2017-12-07
fix(grid-core): allow to place TableColumnVisibility after another table plugins

MaximKudriavtsev:


packages/dx-grid-core/src/plugins/table-column-visibility/computeds.test.js
```diff
@@ -1,18 +1,31 @@
...
+        { type: TABLE_DATA_TYPE, column: { name: 'd' } },
       ];
       const hiddenColumns = ['a', 'c', 'd'];
 
       expect(visibleTableColumns(tableColumns, hiddenColumns))
-        .toEqual([{ column: { name: 'b' } }]);
+        .toEqual([{ type: TABLE_DATA_TYPE, column: { name: 'b' } }]);
+    });
+
+    it('should work with only data type columns', () => {
```
gsobolev: 'should ignore non-data columns' would be more clear

PR #560, opened 2017-12-11
fix(react-grid): fix incorrect applying of several actions in a batch

### LINKED ISSUE #549: Only the last action is executed in a sequence during a batch update
In the following demo, single selection is implemented based on executing several actions: https://stackblitz.com/edit/react-16nodl?file=index.js. These actions are run inside React's batch update. Only the last action is applied. 

### Environment

devextreme-reactive: 1.0.0-beta.1

kvet:
fixes #549


packages/dx-react-grid/src/plugins/grouping-state.jsx
```diff
@@ -36,10 +37,12 @@ export class GroupingState extends React.PureComponent {
       expandedGroups: props.defaultExpandedGroups,
     };
 
+    this.stateHelper = createStateHelper(this);
+
     this.groupByColumn = this.groupByColumn.bind(this);
-    this.toggleGroupExpanded = this.applyReducer.bind(this, toggleExpandedGroups);
-    this.draftGroupingChange = this.applyReducer.bind(this, draftGroupingChange);
-    this.cancelGroupingChange = this.applyReducer.bind(this, cancelGroupingChange);
+    this.toggleGroupExpanded = this.stateHelper.applyReducer.bind(null, toggleExpandedGroups);
```
gsobolev: `.bind(null, /*...*/)` does not look obvious.

kvet:
We will not create component hierarchies. We should prefer composition over inheritance.

Currently, I don't see any problem with this implementations as a createStateHelper receives a delegate in the first parameter. The delegate should have the following shape:

- state getter
- setState method
- getState method
- notifyStateChange method

PR #562, opened 2017-12-11
fix(grid-core): fix problem when grouped columns are hidden

MaximKudriavtsev:


kvet:
We should add the columns Getter to Plugin Developer Reference.

PR #576, opened 2017-12-13
fix(react-grid): remove outline from header cell when mouse click

MaximKudriavtsev:


packages/dx-react-core/src/draggable.jsx
```diff
@@ -6,6 +6,8 @@ import { TouchStrategy } from './draggable/touch-strategy';
 import { MouseStrategy } from './draggable/mouse-strategy';
 import { getSharedEventEmitter } from './draggable/shared-events';
 
+const isDragging = Symbol('isDragging');
```
kvet: Maybe rename it to "draggingHandled"?

PR #578, opened 2017-12-13
docs: hide service properties and components from the reference

gsobolev:


packages/dx-react-grid/docs/reference/table-header-row.md
```diff
@@ -51,7 +51,6 @@ onSort | ({ keepOther: boolean, cancel: boolean }) | An event that changes the a
 allowGroupingByClick | boolean | Specifies whether to render a control that toggles the associated column's grouping state.
 onGroup | () => void | An event that invokes grouping by the associated column.
 allowDragging | boolean | Specifies whether a user can drag a column by the header cell.
-dragPayload | any | A data object that identifies the associated column in the drag-and-drop context.
```
kvet: Should we remove an underlying property too?

packages/dx-react-grid/src/plugins/table-header-row.test.jsx
```diff
@@ -92,15 +93,28 @@ describe('TableHeaderRow', () => {
...
+
     expect(isHeadingTableCell)
       .toBeCalledWith(
-        defaultDeps.template.tableCell.tableRow,
-        defaultDeps.template.tableCell.tableColumn,
+        tableCellProps.tableRow,
+        tableCellProps.tableColumn,
       );
-    expect(tree.find(defaultProps.cellComponent).props())
+    expect(tree.find(defaultProps.cellLayoutComponent).props())
```
kvet: Where are the tests for the Cell component?

packages/dx-react-grid/src/plugins/table-header-row.jsx
```diff
@@ -51,22 +52,24 @@ export class TableHeaderRow extends React.PureComponent {
...
 
                 return (
-                  <HeaderCell
+                  <CellLayout
                     {...params}
-                    column={params.tableColumn.column}
+                    column={column}
+                    draft={draft}
                     getMessage={getMessage}
+                    cellComponent={cellComponent}
```
kvet: Why the Layout component pass rest props to the Cell component? I mean that it is not quite clear why layout should receive the allowSorting property, for example. Maybe we should use the same technique from the GroupingPanel plugin?

packages/dx-react-grid/src/plugins/table-header-row.test.jsx
gsobolev: CellComponent is not created directly by the TableHeaderRow. It is created by TableCellLayout, so all the test cases are there.

kvet: Do this mean that we are not testing the Cell component anymore?

PR #583, opened 2017-12-14
refactor(react-grid): get rid of explicit dragging enabling

kvet:
BREAKING CHANGE:

In order to simplify API, we've decided to remove allowDragging properties in TableHeaderRow and GroupingPanel plugins. Now, to enable dragging it is enough to add the DragDropContext plugin.

### DIFF
--- a/packages/dx-react-grid/src/plugins/table-header-row.jsx
+++ b/packages/dx-react-grid/src/plugins/table-header-row.jsx
@@ -46,7 +46,7 @@ export class TableHeaderRow extends React.PureComponent {
           {params => (
             <TemplateConnector>
               {({
-                sorting, tableColumns, draggingEnabled,
+                sorting, tableColumns, allowDragging,
               }, {
                 setColumnSorting, groupByColumn,
                 changeTableColumnWidths, changeDraftTableColumnWidths,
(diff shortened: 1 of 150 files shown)


packages/dx-react-grid/src/plugins/table-header-row.jsx
```diff
@@ -46,27 +46,26 @@ export class TableHeaderRow extends React.PureComponent {
...
-                sorting, grouping, columns,
+                sorting, tableColumns, draggingEnabled,
               }, {
                 setColumnSorting, groupByColumn,
                 changeTableColumnWidths, changeDraftTableColumnWidths,
               }) => {
                 const { name: columnName } = params.tableColumn.column;
-                const groupingSupported = grouping !== undefined &&
-                    grouping.length < columns.length - 1;
+                const isNotTheLastVisible = tableColumns
```
SergeyAlexeev: isTheLastInvisible?

SergeyAlexeev:
Should we add BC? I mean the case when a user turns off dragging:

```
<TableHeaderRow allowDragging={false} />
```

And doesn't remove `<DragDropContext />`

Should we say: 'To disable table column dragging remove the DragDropContext plugin' ?

kvet:
I've updated the PR description.

packages/dx-react-grid/src/plugins/table-header-row.jsx
kvet: Not sure that I understand how to use the find method here.

SergeyAlexeev:
```js
const atLeastOneVisible = tableColumns.findIndex(({ type }) => type === TABLE_DATA_TYPE) !== -1;
```

kvet: Ok, but I need to check that there are 2 and more columns.

SergeyAlexeev: Oh, sorry, missed this `length > 1`. Anyway, `isNotTheLastVisible` sounds strange)

kvet: I've changed it to `atLeastOneDataColumn`

PR #586, opened 2017-12-15
refactor(react-grid): extract table column extension

kvet:
BREAKING CHANGES:

In order to improve API readability, we extracted the `width` and `align` column properties that the Table and VirtualTable plugins handle into a separate property. Now, the `width` and `align` properties in the Grid `columns` are no longer supported. Use a TableColumnExtension instead:

Before:

```js
const columns= [{ name: 'field', width: 100 }, { name: 'field2' }];

<Grid columns={columns}>
  <Table />
</Grid>
```

After:

```js
const columns= [{ name: 'field1' }, { name: 'field2' }];
const tableColumnExtensions = [{ columnName: 'field1', width: 100 }];

<Grid columns={columns}>
  <Table columnExtensions={tableColumnExtensions} />
</Grid>
```

### DIFF
--- a/packages/dx-grid-core/src/plugins/table/computeds.js
+++ b/packages/dx-grid-core/src/plugins/table/computeds.js
@@ -3,9 +3,10 @@ import { getColumnExtension } from '../../utils/column';
 
 export const tableColumnsWithDataRows = (columns, columnExtensions) =>
   columns.map((column) => {
-    const columnExtension = getColumnExtension(columnExtensions, column.name);
+    const { name } = column;
+    const columnExtension = getColumnExtension(columnExtensions, name);
     return {
-      key: `${TABLE_DATA_TYPE}_${column.name}`,
+      key: `${TABLE_DATA_TYPE}_${name}`,
       type: TABLE_DATA_TYPE,
       width: columnExtension.width,
       align: columnExtension.align,
--- a/packages/dx-grid-core/src/plugins/table/computeds.js
+++ b/packages/dx-grid-core/src/plugins/table/computeds.js
@@ -1,12 +1,17 @@
 import { TABLE_DATA_TYPE, TABLE_NODATA_TYPE } from './constants';
+import { getColumnExtension } from '../../utils/column';
 
-export const tableColumnsWithDataRows = columns =>
-  columns.map(column => ({
-    key: `${TABLE_DATA_TYPE}_${column.name}`,
-    type: TABLE_DATA_TYPE,
-    width: column.width,
-    column,
-  }));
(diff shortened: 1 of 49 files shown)


packages/dx-grid-core/src/plugins/table/computeds.js
```diff
@@ -1,12 +1,17 @@
...
-export const tableColumnsWithDataRows = columns =>
-  columns.map(column => ({
-    key: `${TABLE_DATA_TYPE}_${column.name}`,
-    type: TABLE_DATA_TYPE,
-    width: column.width,
-    column,
-  }));
+export const tableColumnsWithDataRows = (columns, columnExtensions) =>
+  columns.map((column) => {
+    const columnExtension = getColumnExtension(columnExtensions, column.name);
```
SergeyAlexeev: The `column.name` field is used twice and can be extracted.

packages/dx-grid-core/src/plugins/table/computeds.test.js
```diff
@@ -16,12 +16,21 @@ describe('Table Plugin computeds', () => {
...
         .toMatchObject({ width: 100 });
     });
+
+    it('should set align from columnExtension', () => {
+      const columns = [{ name: 'b' }];
+      const columnExtensions = [{ columnName: 'b', align: 'right' }];
+
+      expect(tableColumnsWithDataRows(columns, columnExtensions)[0])
+        .toMatchObject({ align: 'right' });
+    });
```
SergeyAlexeev: Should we add one more test that checks that only the `width` and `align` fields will be applied?

packages/dx-react-demos/src/material-ui/featured-controlled-mode/demo.jsx
```diff
@@ -200,14 +187,16 @@ class DemoBase extends React.PureComponent {
...
+        { name: 'region', title: 'Region' },
+        { name: 'amount', title: 'Sale Amount' },
+        { name: 'discount', title: 'Discount' },
         { name: 'saleDate', title: 'Sale Date' },
         { name: 'customer', title: 'Customer' },
       ],
+      tableColumnExtensions: [
+        { columnName: 'amount', align: 'right', cellComponent: HighlightedCell },
+        { columnName: 'discount', cellComponent: ProgressBarCell },
+      ],
```
SergeyAlexeev: Why there isn't the `width` extension as in the prev version?

packages/dx-grid-core/src/plugins/table/computeds.test.js
kvet: No, I think that these tests are enough

PR #590, opened 2017-12-15
fix(react-grid): fix blinks when reordering header-cell items

MaximKudriavtsev:


packages/dx-react-grid/src/plugins/table-column-reordering.jsx
```diff
@@ -66,7 +71,9 @@ export class TableColumnReordering extends React.PureComponent {
 
     if (relativeSourceColumnIndex === -1) return;
 
-    const cellDimensions = this.getCellDimensions();
+    this.getCellDimensions();
+    const { cellDimensions } = this;
```
kvet: Strange logic. The get* method should return something. Maybe call it cacheCellDimensions?

PR #591, opened 2017-12-15
refactor(react-grid): specify typed columns directly in DataTypeProvider

kvet:
BREAKING CHANGES:

In order to simplify API, we changed the way the DataTypeProvider plugin detects columns associated with a type. Now, to associate a custom formatter and editor with a column, pass a column array to the DataTypeProvider's `for` field:

Before:

```js
const columns= [{ name: 'field', dataType: 'boolean' }, { name: 'field2' }];

<Grid columns={columns}>
  <DataTypeProvider type="boolean" />
</Grid>
```

After:

```js
const columns= [{ name: 'field1' }, { name: 'field2' }];
const booleanColumns = ['field1'];

<Grid columns={columns}>
  <DataTypeProvider for={booleanColumns} />
</Grid>
```

### DIFF
--- a/packages/dx-react-grid/src/plugins/data-type-provider.jsx
+++ b/packages/dx-react-grid/src/plugins/data-type-provider.jsx
@@ -6,7 +6,7 @@ import { PluginContainer, Template } from '@devexpress/dx-react-core';
 export class DataTypeProvider extends React.PureComponent {
   render() {
     const {
-      type,
+      for: forColumnNames,
       formatterComponent: Formatter,
       editorComponent: Editor,
     } = this.props;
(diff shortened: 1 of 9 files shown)


packages/dx-react-grid/src/plugins/data-type-provider.jsx
```diff
@@ -6,7 +6,7 @@ import { PluginContainer, Template } from '@devexpress/dx-react-core';
 export class DataTypeProvider extends React.PureComponent {
   render() {
     const {
-      type,
+      for: forColumnNames,
```
SergeyAlexeev: Why not just `columnNames`? The `for` prefix looks confused.

PR #602, opened 2017-12-20
fix(react-grid-material-ui): use theme constant in the table header background

psamim:
This causes problems for different themes in material-ui.


packages/dx-react-grid-material-ui/src/templates/table.jsx
```diff
@@ -12,7 +12,6 @@ const styles = {
     position: 'sticky',
     top: 0,
     zIndex: 1,
-    background: 'white',
```
SergeyAlexeev: Maybe we should use color from the `theme` object? `theme.palette.background.paper` for example.

psamim: Sure that would be better.

PR #603, opened 2017-12-20
refactor(react-grid): switch to column extensions in LocalSorting

kvet:
BREAKING CHANGES:

We removed the `getColumnCompare` property from the LocalSorting plugin to improve API consistency. Now, use the `columnExtension` property to specify individual column sorting options.

Before:

```js
const columns = [{ name: 'field' }, { name: 'field2' }];
const fieldCompare = /* custom sort compare */;
const getColumnCompare = (columnName) => {
  if (name === 'field') return fieldCompare;
};

<Grid columns={columns}>
  <LocalSorting getColumnCompare={getColumnCompare} />
</Grid>
```

After:

```js
const columns= [{ name: 'field' }, { name: 'field2' }];
const fieldCompare = /* custom sort compare */;
const localSortingColumnExtensions = [
  { columnName: 'field', compare: fieldIdentity },
];

<Grid columns={columns}>
  <LocalSorting columnExtensions={localSortingColumnExtensions} />
</Grid>
```


packages/dx-react-grid/src/plugins/local-sorting.jsx
```diff
@@ -1,21 +1,20 @@
...
 
 const pluginDependencies = [
   { pluginName: 'SortingState' },
 ];
 
 export class LocalSorting extends React.PureComponent {
   render() {
-    const { getColumnCompare } = this.props;
+    const { columnExtensions } = this.props;
+    const getColumnCompare = columnName =>
```
SergeyAlexeev: Should we change `getColumnCompare` to `columnCompare`?

kvet: I left the `getColumnCompare` name in order to make it consistent across local processing plugins.

PR #607, opened 2017-12-20
feat(react-grid): add the ColumnChooser plugin

MaximKudriavtsev:
BREAKING CHANGE:

We have integrated the column chooser into the Grid as a plugin to simplify the API. Now, use the column chooser as follows:

```js
<Grid>
  {/* ... */}
  <TableColumnVisibility
    /* props */
  />
  <Toolbar />
  <ColumnChooser />
</Grid>
```

For details, see [Controlling Column Visibility](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/column-visibility/).


packages/dx-react-grid/src/plugins/column-chooser.jsx
```diff
@@ -0,0 +1,96 @@
...
+        pluginName="ColumnChooser"
+        dependencies={pluginDependencies}
+      >
+        <Template name="toolbarContent">
+          <TemplatePlaceholder />
+          <TemplateConnector>
+            {({ columns, hiddenColumns }, { toggleVisibility }) => ([
+              <ToggleButton
+                key="toggleButton"
+                ref={(button) => { this.button = button; }}
```
kvet: Avoid using refs in plugin

MaximKudriavtsev: @kvet If I move `ref` to the `ToggleButton` component. Is this way better?

kvet: Yes. We may use refs in plugin components. But the plugin should remain "pure"

packages/dx-react-grid-bootstrap3/src/templates/column-chooser/toggle-button.jsx
```diff
@@ -0,0 +1,29 @@
...
+    {...restProps}
+  >
+    <i className="glyphicon glyphicon-eye-close" />
+  </button>
+);
+
+ToggleButton.propTypes = {
+  onToggle: PropTypes.func.isRequired,
+  getMessage: PropTypes.func.isRequired,
+  refHandler: PropTypes.func.isRequired,
```
kvet: The name feels strange. It is used for the overlay target. Maybe call it get/setOverlayTarget? And make it work the same way as the getCellDimension method in TableColumnReordering?

packages/dx-react-grid-bootstrap3/src/templates/column-chooser/toggle-button.test.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { ToggleButton } from './toggle-button';
+
+describe('ToggleButton', () => {
+  it('should pass the className prop to the root element', () => {
+    const tree = shallow((
+      <ToggleButton
+        className="custom-class"
+        onToggle={() => {}}
```
kvet: Can we extract defaultProps here like we did in many plugins?

packages/dx-react-grid/docs/reference/column-chooser.md
```diff
@@ -1,21 +1,50 @@
...
+Describes properties passed to a component that renders the column chooser overlay.
+
+A value with the following shape:
+
+Field | Type | Description
+------|------|------------
+visible | boolean | Specifies whether an overlay is visible.
+target | DOM element | The DOM element that is used to set the position of the overlay.
+onHide | () => void | An event that initiates overlay hiding.
+children | Array&lt;ReactElement&gt; | React elements used to render overlay items.
```
kvet: overlay items -> overlay content?

packages/dx-react-grid/docs/reference/column-chooser.md
```diff
@@ -50,11 +77,36 @@ Field | Type | Description
...
+
+## Plugin Developer Reference
+
+### Imports
+
+Name | Plugin | Type | Description
+-----|--------|------|------------
+columns | Getter | Array&lt;[Column](grid.md#column)&gt; | Grid columns.
+hiddenColumns | Getter | Array&lt;[Column](grid.md#column)&gt; | Columns to be hidden.
+toggleVisibility | Action | ({ columnName: string }) => void | Toggles the column's visibility.
```
kvet: We should call it `toggleColumnVisibility`

packages/dx-react-grid/src/plugins/column-chooser.jsx
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { findDOMNode } from 'react-dom';
```
kvet: Do not use the `react-dom` package here.

packages/dx-react-grid/src/plugins/column-chooser.jsx
```diff
@@ -0,0 +1,100 @@
...
+];
+export class ColumnChooser extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      visible: false,
+    };
+
+    this.handleClickButton = this.handleClickButton.bind(this);
```
kvet: `handleToggle`?

packages/dx-react-grid/src/plugins/column-chooser.jsx
```diff
@@ -0,0 +1,100 @@
...
+      <PluginContainer
+        pluginName="ColumnChooser"
+        dependencies={pluginDependencies}
+      >
+        <Template name="toolbarContent">
+          <TemplatePlaceholder />
+          <TemplateConnector>
+            {({ columns, hiddenColumns }, { toggleVisibility }) => ([
+              <ToggleButton
+                key="toggleButton"
```
kvet: Use React.Fragment instead of an array of keyed elements.

packages/dx-react-grid-bootstrap3/src/templates/column-chooser/toggle-button.jsx
SergeyAlexeev: I'm not sure we need to point Overlay within the Button component. Because they are different components.

kvet: Ok. Anyway, don't use the 'Handler' postfix. It is strange in react.

SergeyAlexeev: It's copied from table-reordering-cell.jsx )

packages/dx-react-grid-bootstrap3/src/plugins/column-chooser.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { ColumnChooser as ColumnChooserBase } from '@devexpress/dx-react-grid';
+import { Overlay } from '../templates/column-chooser/overlay';
+import { Container } from '../templates/column-chooser/container';
```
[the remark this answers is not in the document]
MaximKudriavtsev: You mean that `Container` is a strange name for list items?

packages/dx-react-grid-bootstrap3/src/templates/column-chooser/toggle-button.jsx
kvet: But it's an internal name

packages/dx-react-grid/src/plugins/column-chooser.jsx
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    this.handleToggle = this.handleToggle.bind(this);
+    this.handleHide = this.handleHide.bind(this);
+    this.getButtonRef = this.getButtonRef.bind(this);
+  }
+  getButtonRef(button) {
+    this.button = button;
+  }
+  handleToggle() {
```
MaximKudriavtsev: Maybe we must change this name: handleToggle -> handleOpen. Because it is not toggle function.
We will have toggleOpen and toggleHide

packages/dx-react-grid-bootstrap3/src/plugins/column-chooser.jsx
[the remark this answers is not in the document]
SergeyAlexeev: Actually, we don't have `list items`. We have `items` that are placed within Container. Yes, in MUI we use the ListItem component to render column chooser item, but it's only for UI representation. Column chooser items can be rendered as buttons placed side by side or something like this. So, I'd prefer to keep `Container` as a component name. It's a quite abstract.

packages/dx-react-grid/docs/reference/column-chooser.md
```diff
@@ -1,21 +1,51 @@
...
+
+Describes properties passed to a component that renders the column chooser toggle button.
+
+A value with the following shape:
+
+Field | Type | Description
+------|------|------------
+onToggle | () => void | An event that initiates overlay showing or hiding.
+getMessage | ([messageKey](#localization-messages): string) => string | Returns title for the column chooser toggle button. Available in the "@devexpress/dx-react-grid-material-ui" package.
+buttonRef | (node) => void | A function that take a React element of the root element.
```
MaximKudriavtsev: @kvet what will you say about this description?
Maybe (React element) => void ?

kvet: Let's define it like this: `(ref: ReactElement) => void`

packages/dx-react-grid/docs/reference/column-chooser.md
```diff
@@ -1,21 +1,51 @@
...
+### ColumnChooserOverlayProps
+
+Describes properties passed to a component that renders the column chooser overlay.
+
+A value with the following shape:
+
+Field | Type | Description
+------|------|------------
+visible | boolean | Specifies whether an overlay is visible.
+target | React element | The React element that is used to set the position of the overlay.
```
kvet: `ReactComponent`

packages/dx-react-grid-material-ui/src/templates/column-chooser/overlay.jsx
```diff
@@ -0,0 +1,32 @@
...
+export const Overlay = ({
+  visible, onHide,
+  children, target,
+  ...restProps
+}) => (
+  <Popover
+    open={visible}
+    anchorEl={target}
+    onRequestClose={onHide}
+    anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
```
kvet: Maybe we should position it at center?

packages/dx-react-grid-material-ui/src/templates/column-chooser/toggle-button.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+export const ToggleButton = ({
+  onToggle,
+  getMessage,
+  buttonRef,
+  ...restProps
+}) => (
+  <Tooltip
+    title={getMessage('hiddenColumns')}
+    placement="bottom-end"
```
kvet: The same here

packages/dx-react-grid-material-ui/src/templates/column-chooser/overlay.jsx
SergeyAlexeev: The horizontal alignment is quite [good](https://storage.googleapis.com/material-design/publish/material_v_12/assets/0Bzhp5Z4wHba3WVAwRkdSdURJZWc/component-menus-usage-appbar-dropdown-xhdpi-004.webm). But not vertical)

packages/dx-react-grid/docs/reference/column-chooser.md
```diff
@@ -29,22 +29,22 @@ A value with the following shape:
...
+Describes properties passed to a component that renders a button that invokes the column chooser.
 
 A value with the following shape:
 
 Field | Type | Description
 ------|------|------------
 onToggle | () => void | An event that initiates overlay showing or hiding.
-getMessage | ([messageKey](#localization-messages): string) => string | Returns title for the column chooser toggle button. Available in the "@devexpress/dx-react-grid-material-ui" package.
-buttonRef | (ref: ReactElement) => void | A function that take a React element of the root element.
+getMessage | ([messageKey](#localization-messages): string) => string | Returns the toggle button's text. Available in the "@devexpress/dx-react-grid-material-ui" package.
```
kvet: Why do we write that the getMessage function that receives the message key argument returns only the single message? It feels strange.

viterobk: Looks like it should sound like "Returns a specified localization message."

PR #609, opened 2017-12-21
refactor(react-grid): switch to column extensions in TableGroupRow

kvet:
BREAKING CHANGES:

We extracted the column's `showWhenGrouped` property that the TableGroupRow plugin handles to a column extension to improve the API readability.
The TableGroupRow plugin's `showColumnsWhenGrouped` property type has been changed to `boolean` and this property affects all columns. If you need to control specific columns' visibility, use column extensions.

Before:

```js
const columns = [{ name: 'field', showWhenGrouped: true }, { name: 'field2' }];

<Grid columns={columns}>
  <TableGroupRow />
</Grid>
```

After:

```js
const columns= [{ name: 'field' }, { name: 'field2' }];
const tableGroupColumnExtensions = [
  { columnName: 'field', showWhenGrouped: true },
];

<Grid columns={columns}>
  <TableGroupRow columnExtensions={tableGroupColumnExtensions} />
</Grid>
```

Note that column extension properties have higher priority than corresponding plugin properties.


packages/dx-react-grid/docs/guides/grouping.md
```diff
@@ -73,8 +73,8 @@ You can also assign a function that returns a Boolean value depending on the `co
     />
     <LocalGrouping />
     <Table />
-    <TableHeaderRow showColumnWhenGrouped={columnName => columnName === 'city' || columnName === 'car'}/>
-    <TableGroupRow />
+    <TableHeaderRow />
+    <TableGroupRow showColumnWhenGrouped />
```
kvet: Should we change the name to `showColumnsWhenGrouped`?

SergeyAlexeev: Makes sense.

kvet: @viterobk, can you help us?

Issue #288, opened 2017-08-28
Custom CSS Classes on rendered table


terndrup:
Hey guys,

sorry if this is obvious and I just missed it but I've been looking through the Docs and Demos and unable to find a place where I can declare custom CSS classes that the grid should add to the table HTML element it creates.

This is relevant in for example bootstrap where you may want to use some of the bootstrap built-in table classes such as ``` table-bordered ``` ``` table-striped ``` ``` table-condensed ``` etc.

I tried adding them using the className react decorator on TableView or Grid elements without success.

Any ideas?

kvet:
Hi,
 
Thank you for reporting this scenario. This is a very common use case. We have no API for adding custom CSS classes at this moment. We will provide a table customization API in future releases.
 
In the meantime, we suggest using the following workaround: https://stackblitz.com/edit/react-3kavn7?file=index.js

SergeyAlexeev:
I've created a separate [issue](https://github.com/DevExpress/devextreme-reactive/issues/627) that describes a similar scenario.

PR #610, opened 2017-12-21
pass className in Table

caub:
could help for #288


SergeyAlexeev:
Hi, thank you for sharing your idea!
In the [latest](https://github.com/DevExpress/devextreme-reactive/blob/master/CHANGELOG.md#features) release, we introduced plugins components that allow customizing Grid elements, such as rows, cells, panels, etc. I think we should go the same way to customize the 'table' element. I mean extract a separate component for it. In this case, a user will be able to add any property to the table, not only className.

SergeyAlexeev:
I've created a separate [issue](https://github.com/DevExpress/devextreme-reactive/issues/627) that describes a similar scenario.

PR #617, opened 2017-12-25
refactor(react-grid): give column properties greater precedence

kvet:
BREAKING CHANGE:

A column's `getCellValue` function has gotten higher priority than that defined in the Grid component.


packages/dx-grid-core/src/plugins/grid-core/computeds.test.js
```diff
@@ -35,21 +35,19 @@ describe('GridCore Plugin computeds', () => {
...
-      expect(cellValueGetter(getCellValue, columns)({ a: 1 }, 'a'))
+      expect(cellValueGetter(undefined, columns)({ a: 1 }, 'a'))
         .toEqual(1);
-      expect(cellValueGetter(getCellValue, columns)({ b: 2 }, 'b'))
+      expect(cellValueGetter(undefined, columns)({ b: 2 }, 'b'))
         .toEqual(2);
     });
 
     it('should work with column defined cell access', () => {
       const columns = [{ name: 'a', getCellValue: row => row.b }, { name: 'b' }];
```
SergeyAlexeev: Column name is `a` but we get value by `b`. It's a bit confusing.

packages/dx-grid-core/src/plugins/grid-core/computeds.test.js
```diff
@@ -35,21 +35,19 @@ describe('GridCore Plugin computeds', () => {
...
-      const getCellValue = null;
 
-      expect(cellValueGetter(getCellValue, columns)({ b: 1 }, 'a'))
+      expect(cellValueGetter(undefined, columns)({ b: 1 }, 'a'))
         .toEqual(1);
-      expect(cellValueGetter(getCellValue, columns)({ b: 2 }, 'b'))
+      expect(cellValueGetter(undefined, columns)({ b: 2 }, 'b'))
         .toEqual(2);
     });
```
SergeyAlexeev: I think we should add a test case that checks that `column.getCellValue` has greater priority than `Grid.getCellValue` one.

```js
const columns = [{ name: 'a', getCellValue: row => row.a }, { name: 'b' }];
const getCellValue = (row, columnName) => row[`_${columnName}`];

expect(cellValueGetter(getCellValue, columns)({ a: 1, _a: 2 }, 'a'))
   .toEqual(1);
```

packages/dx-grid-core/src/plugins/grid-core/computeds.test.js
gsobolev: We should not write confusing tests anyway.

packages/dx-react-grid/docs/guides/data-accessors.md
```diff
@@ -74,8 +74,6 @@ The following demo shows this approach in action:
 
 .embedded-demo(data-accessors/custom-data-accessors)
 
-Note that the Grid's `getCellValue` property has a higher priority than the column's property.
```
gsobolev: We have changed the priorities. Why do we remove the note completely instead of changing it?

kvet: We haven't such note in other places. We don't need to leave it here for now.

PR #618, opened 2017-12-26
refactor(react-grid): rename properties from allow* to *Enabled

kvet:
BREAKING CHANGES:

The TableHeaderCellProps interface's `allowResizing` property has been renamed to `resizingEnabled` to improve the API consistency.


SergeyAlexeev:
BC doesn't describe all renames.

kvet:
The BC text is correct

PR #640, opened 2018-01-03
fix(react-grid-material-ui): import 'VisibilityOff' icon from separat…

helsus:
Because everyone add `node_modules` to `exclude` option of `babel-loader` when using webpack (So babel plugins can't transform this code to prevent full import)

I've tested this by changes in my project `node_modules`:

from:
```javascript
import { VisibilityOff } from 'material-ui-icons';
```
to:
```javascript
import VisibilityOff from 'material-ui-icons/VisibilityOff';
```
And my bundle have been decreased by ~400kb.

I'm not sure, but perhaps there is a better way to fix tree-shaking?

### DIFF
--- a/packages/dx-react-grid-material-ui/src/templates/column-chooser/toggle-button.jsx
+++ b/packages/dx-react-grid-material-ui/src/templates/column-chooser/toggle-button.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { IconButton, Tooltip } from 'material-ui';
-import { VisibilityOff } from 'material-ui-icons';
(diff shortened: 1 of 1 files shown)


kvet:
Hi,

Thank you for your cooperation. Currently, we don't see a better way to fix it.

Maher4Ever:
When can we expect a release including this change? I'd love to have as this issue currently accounts for more than 50% of the size of my website according to `source-map-explorer`!

kvet:
Hi @Maher4Ever,

We've released the [1.0.0-rc.1](https://github.com/DevExpress/devextreme-reactive/releases/tag/v1.0.0-rc.1) version containing described changes.

Maher4Ever:
@kvet Awesome! Thank you guys for this amazing project

PR #644, opened 2018-01-09
docs: embed demo sources

kvet:


site/_layouts/default.html
```diff
@@ -64,7 +64,7 @@
                 </div>
                 <div class="col-md-9 copyright">
                   <p>
-                    Copyright © 2011-2017 Developer Express Inc.</br>
+                    Copyright © 2011-2018 Developer Express Inc.</br>
```
SergeyAlexeev: Can we change year automatically, like `{{ site.time | date: '%Y' }}`?

kvet: Yes, we should

SergeyAlexeev:
Can't build demos(`yarn build:site`). [Error](https://www.screencast.com/t/6TBA7P0y).

PR #645, opened 2018-01-09
refactor(react-grid): use the 'state' name as a parameter in toggle actions

kvet:
BREAKING CHANGE:

The `toggleSelection` action's `selected` parameter has been renamed to `state`.


packages/dx-grid-core/src/plugins/selection-state/reducers.test.js
```diff
@@ -1,92 +1,92 @@
...
       const selection = [1, 2, 3];
       const payload = { rowIds: [3, 4] };
 
-      const nextSelection = setRowsSelection(selection, payload);
+      const nextSelection = toggleSelection(selection, payload);
       expect(nextSelection).toEqual([1, 2, 3, 4]);
     });
 
-    it('should remove from selection all rows is selected', () => {
+    it('should remove from selection all rows is state', () => {
```
SergeyAlexeev: The test name sounds strange

PR #663, opened 2018-01-10
fix(react-grid): set higher precedence to props in state plugins

### LINKED ISSUE #662: Issue with remote data, filtering and pagination
### Current Behaviour
When using remote data + filtering + pagination, if you are in page 17 and filter for something that only has 2 pages, grid currentPage stays in 17 and don't show any data.

If you change currentPage state to 0 when changingFilter, the grid shows data, but pagination doesn't work as intented (Ex: In my case, pagination shows 321 of 26)

### Expected Behaviour
When using remote data + filtering + pagination, if you are in page 17 and filter for something that only has 2 pages, grid currentPage should change to first page and fix pagination.

1- Create a grid with 2 or more pages.
2- Go to page >= 2
3- Filter for something that is only showed in pages before the page selected in step 2.
4- Grid doesn't show data.

### Your Environment
devextreme-reactive: 1.0.0-beta.3
react: 16.2.0
browser: Chrome
bootstrap: none
react-bootstrap: none
material-ui: 1.0.0-beta.27

kvet:
fixes #662


packages/dx-react-grid/src/plugins/table-column-reordering.jsx
```diff
@@ -28,7 +28,7 @@ export class TableColumnReordering extends React.PureComponent {
     super(props);
 
     this.state = {
-      order: props.defaultOrder || props.order,
+      order: props.defaultOrder,
```
MaximKudriavtsev: missing [ ] in defaultProps

PR #670, opened 2018-01-12
fix(react-grid): display empty message when all data columns are hidden

MaximKudriavtsev:


packages/dx-grid-core/src/plugins/table-column-visibility/helpers.js
```diff
@@ -0,0 +1,20 @@
+export const isEmptyMessageShow = (grouping, hiddenColumnNames, columns) => {
+  if (grouping === undefined) return false;
+  if (columns.length === hiddenColumnNames.length) return true;
+
+  const groupingColumnNames = grouping.map(group => group.columnName);
+  const showedColumnNames = columns.reduce((acc, showedColumn) => {
```
SergeyAlexeev: visibleColumnNames?

SergeyAlexeev: Can we use `tableColumns` instead of such calculation?

packages/dx-react-grid/src/plugins/table-column-visibility.jsx
```diff
@@ -65,7 +65,8 @@ export class TableColumnVisibility extends React.PureComponent {
         <Template name="table">
           {params => (
             <TemplateConnector>
-              {({ columns }) => (columns.length === hiddenColumns.length
+              {({ tableColumns }) =>
+                (isEmptyMessageShow(tableColumns)
```
SergeyAlexeev: Do we really need the separated `isEmptyMessageShow` helper? It doesn't contain any complex logic and is used only once.

MaximKudriavtsev: For simple testing this functionality.

packages/dx-grid-core/src/plugins/table-column-visibility/helpers.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+describe('TableColumnVisibility helpers', () => {
+  describe('#isEmptyMessageShow', () => {
+    it('should show empty message when all columns are hidden', () => {
+      const grouping = [
+        { columnName: 'a' },
+        { columnName: 'b' },
+      ];
+
+      expect(isEmptyMessageShow(grouping, []))
```
SergeyAlexeev: `isEmptyMessageShow` takes only one argument - `tableColumns`. But here there are two args. And one of them is called `grouping`.

And let's rename `isEmptyMessageShow`. `showEmptyMessage` or `tableColumnsExist`.

packages/dx-grid-core/src/plugins/table-column-visibility/helpers.js
```diff
@@ -0,0 +1,4 @@
+import { TABLE_DATA_TYPE } from '../table/constants';
+
+export const isEmptyMessageShow = tableColumns =>
```
kvet: Maybe use the 'isEmptyMessageShown' name here?

packages/dx-grid-core/src/plugins/table-column-visibility/helpers.test.js
```diff
@@ -0,0 +1,31 @@
...
+    it('should show empty message when all showed columns are grouped', () => {
+      const tableColumns = [
+        { type: 'any', column: { name: 'a' } },
+      ];
+
+      expect(isEmptyMessageShow(tableColumns))
+        .toBeTruthy();
+    });
+
+    it('should not show empty message while a showed column is not grouped', () => {
```
kvet: shown?

packages/dx-grid-core/src/plugins/table-column-visibility/helpers.js
SergeyAlexeev: I'd prefer something like `tableDataColumnsExist`. In this case, we can use it as follows:
```js
<TemplateConnector>
  {({ tableColumns }) =>
    (tableDataColumnsExist(tableColumns)
    ? <TemplatePlaceholder />
    : <EmptyMessage
      getMessage={getMessage}
      {...params}
    />
  )}
</TemplateConnector>
```

packages/dx-grid-core/src/plugins/table-column-visibility/helpers.test.js
```diff
@@ -0,0 +1,31 @@
+import { TABLE_DATA_TYPE } from '../table/constants';
+import { tableDataColumnsExist } from './helpers';
+
+describe('TableColumnVisibility helpers', () => {
+  describe('#tableDataColumnsExist', () => {
+    it('should show empty message when all columns are hidden', () => {
```
gsobolev: Does the `tableDataColumnsExist` method really show any messages?
The same is for all other tests in this file.

PR #674, opened 2018-01-16
docs: rename and sort guides

viterobk:


site/_data/docs/navigation.yml
```diff
@@ -6,16 +6,14 @@ react:
           path: /react/grid/docs/guides/getting-started
         - title: Plugin Overview
           path: /react/grid/docs/guides/plugin-overview
+        - title: Fundamentals
+          path: /react/grid/docs/guides/grid-basics
```
MaximKudriavtsev: Maybe we should rename filenames too? Because after some time we become entangled in the notation

Issue #671, opened 2018-01-14
Server side rendering with devextreme and material ui?


kkotwal94:
After refreshing the page (and going through my ssr) it looks like none of the css is sustained unless I navigate through my app and get back to it. Are there any examples of how this is done correctly?

Im using the exact same code from the controlled 'react material ui grid' example:

https://github.com/kkotwal94/DrivingService <- (in develop branch) where the component is under components / demoGrid, and the SSR is under server/render/pageRender.jsx. I use the material ui example for how this is done. I utilize demo grid in Students.jsx.

Here is a pic of what happens post refresh: 
![image](https://user-images.githubusercontent.com/8924340/34913058-97d86032-f8c0-11e7-8d6a-c452213f08bb.png)

Everything else renders fine (all other pages) in production mode and dev mode. I have no clue what im missing here.

SergeyAlexeev:
Hi,
 
Thank you for your interest in our product! I just tested how the React Grid works with Next.js and Material UI and it all works fine as described  [here](https://github.com/mui-org/material-ui/tree/v1-beta/examples/nextjs). I could not find any problems or UI issues.

kkotwal94:
Hmm maybe I can produce a working example and host it later today, I'm not sure what i'm doing differently. But after looking around some more I realized most of my material-ui css is not syncing right (jss tags are mismatched with SSR on refresh). I downgraded back to react 15 and all seems good, however I can't use the grid on 15. I will do some more experimentation later today.

SergeyAlexeev:
Yes, you are right.

kkotwal94:
http://transportation.kkotwal.me/
I hosted it, if you click on login  you can log in with yea@yea.com, password: 123, or you can just sign up where the username has to be a email it doesnt matter. After wards if you navigate to the students button on the navigation (if you click on transportation tracker after logging in you should be back to the root page / view). You will see the dev extreme controlled grid example. If you hit refresh on that page you will see all the css is messed up. In case you arent sure what the page is: http://transportation.kkotwal.me/students. The source is here: https://github.com/kkotwal94/DrivingService/tree/UpdateReact . The server side rendering is located https://github.com/kkotwal94/DrivingService/tree/UpdateReact/server/render. The component for the devExtreme component is called DemoBase.jsx in the components folder, and the container that renders this is https://github.com/kkotwal94/DrivingService/blob/UpdateReact/app/containers/students/Students.jsx.

[...]

PR #684, opened 2018-01-18
fix(react-grid-material-ui): use long paths for modules

kvet:
related to #671


packages/dx-react-demos/src/material-ui/basic/table-cell-template.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { TableCell, Paper } from 'material-ui';
+import Paper from 'material-ui/Paper';
+import { TableCell } from 'material-ui/Table';
```
SergeyAlexeev: Can we remove curly braces?

oliviertassinari:
Good call :+1:.

PR #694, opened 2018-01-24
fix(react-core): fixes crash when using with webpack-dev-server and Hot Module Reloading

tpisto:


kvet:
Hi,

Thank you for your cooperation! I read the changes and cannot agree with them. The position property in the Getter, Action, and Template contains vital information. It can't be replaced with a zero.

I think that the main problem is in the following code: https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-core/src/plugged/plugin-indexer.jsx#L22-L24. The problem is also described here: https://github.com/gaearon/react-hot-loader/tree/next#checking-element-types.

As far as I understand, the fix should change the identification method of the Getter, Action, and Template. I also want to mention that we don't plan to add a dependency to the `react-hot-loader` package. As a possible fix, we can assign a Symbol to the Getter, Action, and Template and check the type by that Symbol. Something like this:

```jsx
const indexableComponent = Symbol();

// define
Getter[indexableComponent] = true;

// check
element.type[indexableComponent] === true
```

tpisto:
Thank you for your quick reply - and also thank you for creating this great grid component that has great potential to become the de-facto grid for react!

I think your solution looks great and the analysis is correct. Please could you incorporate it to next release so that this component wouldn't crash in hot-reload scenarios? I think most React devs are using hot-reload... Thank you again!

packages/dx-react-core/src/plugged/plugin-indexer.jsx
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Getter } from './getter';
-import { Action } from './action';
-import { Template } from './template';
+
+export const indexableComponent = Symbol('indexableComponent');
```
kvet: Change constant name to `INDEXABLE_COMPONENT`

PR #696, opened 2018-01-25
chore: fix spelling of a Select All

MaximKudriavtsev:


packages/dx-react-grid/docs/guides/selection.md
```diff
@@ -30,7 +30,7 @@ In some scenarios, it is useful to highlight selected rows instead of using chec
 
 ## Select All
 
-Perform the following steps to enable the "select/deselect all" capability:
+Perform the following steps to enable the select/deselect all capability:
```
RomanTsukanov: this should be "the Select All capability"

PR #704, opened 2018-01-29
refactor(react-demos): remove propTypes from all demos

MaximKudriavtsev:


packages/dx-react-demos/src/demo-viewer/theme-viewer.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint react/prop-types: "error" */
```
kvet: Can we add an additional `.eslintrc` for that folder?

PR #707, opened 2018-01-30
chore: update rollup on all packages

MaximKudriavtsev:


packages/dx-react-grid-material-ui/src/templates/empty-message.jsx
```diff
@@ -1,16 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Toolbar from 'material-ui/Toolbar';
+import ToolbarMUI from 'material-ui/Toolbar';
```
kvet: This is an implicit fix for rollup's bug.

PR #713, opened 2018-02-01
refactor(react-grid): rename expanded details getter

kvet:
BREAKIN CHANGES:

The `expandedRowIds` getter of the RowDetailState plugin has been renamed to `expandedDetailRowIds`.


packages/dx-react-grid/docs/reference/row-detail-state.md
```diff
@@ -27,4 +27,4 @@ none
 Name | Plugin | Type | Description
 -----|--------|------|------------
 toggleDetailRowExpanded | Action | ({ rowId }) => void | Expands/collapses the specified row.
-expandedRowIds | Getter | Array&lt;number &#124; string&gt; | Currently expanded rows.
+expandedDetailRowIds | Getter | Array&lt;number &#124; string&gt; | Currently expanded rows.
```
SergeyAlexeev: > Currently expanded rows.

Currently expanded row ids ?

kvet: I'd prefer not to change the description. As it will lead to full recheck for several days.

SergeyAlexeev: Ok)

PR #716, opened 2018-02-01
chore: update deps

MaximKudriavtsev:


packages/dx-react-grid-material-ui/src/templates/paging-panel/pager.jsx
```diff
@@ -4,13 +4,11 @@ import classNames from 'classnames';
 import { withStyles } from 'material-ui/styles';
 import { PageSizeSelector } from './page-size-selector';
 import { Pagination } from './pagination';
-import { getBorderColor } from '../utils';
 
 const styles = theme => ({
   pager: {
     overflow: 'hidden',
     padding: theme.spacing.unit * 1.5,
-    borderTop: getBorderColor(theme),
```
MaximKudriavtsev: Because this border is 2 px

gsobolev: This leads to the PaginPanel top border disappearance (see the featured-controlled-mode demo)

packages/dx-react-demos/src/demo-sources/grid-data-types/material-ui/editors.jsx
```diff
@@ -31,7 +31,7 @@ const BooleanEditor = ({ value, onValueChange }) => (
     input={<Input />}
     value={value ? 'Yes' : 'No'}
     onChange={event => onValueChange(event.target.value === 'Yes')}
-    style={{ width: '100%' }}
+    style={{ width: '100%', marginTop: '4px' }}
```
gsobolev: Is not it possible to calculate this magic '4px' value somehow?

MaximKudriavtsev: It's mui bug, and we add this style because don't have `const theme` here

SergeyAlexeev: Yes, it is. It's kind of 'fix after update MUI':) So, we decided to keep code simple and not to use `withStyles` etc.

PR #728, opened 2018-02-06
fix(react-grid): allow adding custom styles to some components

gsobolev:


packages/dx-react-grid-bootstrap3/src/templates/table-container.jsx
```diff
@@ -1,12 +1,12 @@
...
 import PropTypes from 'prop-types';
 
-export const TableContainer = ({ children, ...restProps }) => (
+export const TableContainer = ({ children, style, ...restProps }) => (
   <div
     className="table-responsive"
     {...restProps}
     style={{
-      ...restProps.style,
+      ...style,
```
SergeyAlexeev: Where should we spread user styles? Sometimes we put them before our internal ones, sometimes after.

PR #737, opened 2018-02-09
chore: fix uglify js crash

kvet:


packages/dx-react-demos/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
       "sourceMap": true,
       "module": "es6",
-      "target": "es6",
+      "target": "es5",
+      "lib": ["es5", "es6", "dom"],
```
gsobolev: Why do we need "es5" here?

PR #749, opened 2018-02-15
fix(react-grid): calculate visible rows when rendering VirtualTable

### LINKED ISSUE #748: VirtualTable should update when height changes

### Current Behaviour

Changing the height of VirtualTable doesn't update its contents.
Rows hidden because of a small height stay hidden when the height is increased.
After increasing the height scrolling is necessary to update the view.

### Expected Behaviour

Scrolling should not be necessary. Changing the height should suffice.

https://codesandbox.io/s/1yor2kwvkq

### Your Environment

devextreme-reactive: 1.0.0
react: 16.2.0
browser: Chrome 64.0.3282.167
bootstrap: none
react-bootstrap: none
material-ui: 1.0.0-beta.31

kvet:
fixes #748

### DIFF
--- a/packages/dx-react-grid/src/components/table-layout/virtual-table-layout.jsx
+++ b/packages/dx-react-grid/src/components/table-layout/virtual-table-layout.jsx
@@ -13,37 +13,35 @@ export class VirtualTableLayout extends React.PureComponent {
     super(props);
 
     this.state = {
-      visibleBodyRows: [],
+      rowHeights: new Map(),
+      viewportTop: 0,
     };
 
     this.rowRefs = new Map();
-    this.rowHeights = new Map();
-    this.viewportTop = 0;
     this.updateViewport = this.updateViewport.bind(this);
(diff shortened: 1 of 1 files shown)


packages/dx-react-grid/src/components/table-layout/virtual-table-layout.jsx
```diff
@@ -104,9 +102,8 @@ export class VirtualTableLayout extends React.PureComponent {
     }
 
     if (this.viewportTop !== node.scrollTop) {
-      this.viewportTop = node.scrollTop;
       this.setState({
-        visibleBodyRows: this.getVisibleRows(),
+        viewportTop: node.scrollTop,
```
SergeyAlexeev: Can we extract `scrollTop` from `node`? It's using a lot of times.

kvet: Yes, we can, but I am not sure that we should. Seems like function will become more verbose.

PR #784, opened 2018-02-28
fix(react-grid-material-ui): update material-ui and material-ui-icons version to 1.0.0-beta.35

kvet:


SergeyAlexeev:
Default color is pink :)

image

kvet:
BC from https://github.com/mui-org/material-ui/releases/tag/v1.0.0-beta.34. I'm not sure that we should specify `color="primary"`.

PR #817, opened 2018-03-06
chore: add npm version badge

kvet:


MaximKudriavtsev:
Do we add a badge with `@next` version?
I think it displays for customers what current beta version.
Also, add to docs this information.

kvet:
I am not sure that we should add this information here. This information is not typically added to the readme.

PR #821, opened 2018-03-06
fix(react-demos): get rid of warnings in the Featured Controlled demos

gsobolev:


packages/dx-react-demos/src/theme-sources/material-ui/components/integer-type-provider.jsx
```diff
@@ -0,0 +1,45 @@
...
+const IntegerEditorBase = ({ value, onValueChange, classes }) => (
+  <Input
+    type="number"
+    classes={{
+      root: classes.inputRoot,
+      input: classes.numericInput,
+    }}
+    value={value}
+    inputProps={{
+      step: 100,
```
SergeyAlexeev: It looks like step is too big :)

gsobolev: The 'amount' field value does not surpass the lower edge of 1000, so I thought it is OK to use 100 as a step. What value do you suggest instead?

kvet: You can remove step at all.

PR #834, opened 2018-03-13
add option to have title as object

RahavLussato:
i want to be able to create the title as object to have it as div with tooltip node:
```
<div>
         <span data-tip data-for={`test`}>{title}</span>
         <ReactTooltip id={`test`} effect="solid">
                <div style={{lineHeight:1.5}}>
                     <span>Resolution: {resolution}</span>
                     <br/>
                     <span>Name: {title}</span>
                </div>
         </ReactTooltip>
</div>
```


kvet:
Hi,

Thank you for your desire to make our products better. I see the problem you are trying to solve. Although it could be a good notification about limited customization options, we can't merge these changes, because we have plans to improve customization experience in a different way. So, I am closing the PR.

PR #837, opened 2018-03-14
feat(react-grid): add minColumnWidth option

Krijovnick:


packages/dx-react-grid/docs/reference/table-column-resizing.md
```diff
@@ -13,6 +13,7 @@ A plugin that manages table column widths.
 Name | Type | Default | Description
 -----|------|---------|------------
 columnWidths? | Array&lt;[TableColumnWidthInfo](#tablecolumnwidthinfo)&gt; | | Specifies column widths.
+minColumnWidth? | number | `45` for [Bootstrap3](https://www.npmjs.com/package/@devexpress/dx-react-grid-bootstrap3); `55` for [Bootstrap4](https://www.npmjs.com/package/@devexpress/dx-react-grid-bootstrap4); `40` for [Material UI](https://www.npmjs.com/package/@devexpress/dx-react-grid-material-ui); | Specifies min column width.
```
gsobolev: I believe it's better to use 'minimal' instead of 'min' in the description.

packages/dx-react-grid-bootstrap4/src/plugins/table-column-resizing.jsx
```diff
@@ -1,12 +1,23 @@
 import * as React from 'react';
+import * as PropTypes from 'prop-types';
 import { TableColumnResizing as TableColumnResizingBase } from '@devexpress/dx-react-grid';
 
 export class TableColumnResizing extends React.PureComponent {
   render() {
+    const { minColumnWidth } = this.props;
```
SergeyAlexeev: const { minColumnWidth, ...restProps } = this.props;

packages/dx-react-grid-material-ui/src/plugins/table-column-resizing.jsx
```diff
@@ -1,12 +1,23 @@
 import * as React from 'react';
+import * as PropTypes from 'prop-types';
 import { TableColumnResizing as TableColumnResizingBase } from '@devexpress/dx-react-grid';
 
 export class TableColumnResizing extends React.PureComponent {
   render() {
+    const { minColumnWidth } = this.props;
```
SergeyAlexeev: The same here.

packages/dx-react-grid/docs/reference/table-column-resizing.md
SergeyAlexeev: Should we mention units here? I mean `40px` etc.

Krijovnick: @SergeyAlexeev we didn't mention units before

SergeyAlexeev:
Just a question. Should we introduce column extensions like:

```
<TableColumnResizing
   columnExtensions={[
      { columnName: 'name', minWidth: 100 }
   ]}
/>
```

?

Of course, it can be another PR.

Krijovnick:
@SergeyAlexeev I think we should get a request before adding this option to the column extensions. Perhaps minBarWidth option will be enough.

packages/dx-react-grid/docs/reference/table-column-resizing.md
```diff
@@ -13,6 +13,7 @@ A plugin that manages table column widths.
 Name | Type | Default | Description
 -----|------|---------|------------
 columnWidths? | Array&lt;[TableColumnWidthInfo](#tablecolumnwidthinfo)&gt; | | Specifies column widths.
+minColumnWidth? | number | `45` for [Bootstrap3](https://www.npmjs.com/package/@devexpress/dx-react-grid-bootstrap3); `55` for [Bootstrap4](https://www.npmjs.com/package/@devexpress/dx-react-grid-bootstrap4); `40` for [Material UI](https://www.npmjs.com/package/@devexpress/dx-react-grid-material-ui); | Specifies minimal column width.
```
MaximKudriavtsev: Maybe we should change property name to `minColumnsWidth`? It shows that minimal width defines for each column.

Krijovnick: @MaximKudriavtsev, in this case, it is pretty clear that 'minColumnWidth' option is used for all columns. in addition, other our properties use the singular form and defined for all columns/rows.

packages/dx-react-grid/src/plugins/table-column-resizing.jsx
```diff
@@ -69,6 +71,7 @@ TableColumnResizing.propTypes = {
   defaultColumnWidths: PropTypes.array,
   columnWidths: PropTypes.array,
   onColumnWidthsChange: PropTypes.func,
+  minColumnWidth: PropTypes.number.isRequired,
```
rememberlenny: Was `isRequired` intentional here?

PR #868, opened 2018-03-20
docs: add static column width demo to guides

gsobolev:


SergeyAlexeev:
Do we expect horizontal scrollbar in a demo?

packages/dx-react-demos/src/demo-sources/grid-basic/static-column-width.jsxt
```diff
@@ -0,0 +1,51 @@
...
+        { name: 'product', title: 'Product' },
+        { name: 'amount', title: 'Sale Amount' },
+      ],
+      rows: generateRows({ columnValues: globalSalesValues, length: 8 }),
+      tableColumnExtensions: [
+        { columnName: 'region', width: 180 },
+        { columnName: 'sector', width: 130 },
+        { columnName: 'customer', width: 240 },
+        { columnName: 'product', width: 210 },
+        { columnName: 'amount', width: 140 },
```
MaximKudriavtsev: Maybe we should display the `align` property too?

gsobolev:
I don't think horizontal scrollbar is a problem here. It may also emphasize that columns preserve widths specified in the config.

packages/dx-react-demos/src/demo-sources/grid-basic/static-column-width.jsxt
gsobolev: I believe we should not mix two demos

PR #873, opened 2018-03-21
docs: add column alignment demo to guides

gsobolev:


MaximKudriavtsev:
Maybe we should create only one demo for alignment and widths?

PR #888, opened 2018-03-28
[wip]feat(react-grid): add highlight to text on use the Searching feature

MaximKudriavtsev:


FGCorp:
is there an example of this to use on table.i cant figure it out.

FGCorp:
i checked the example its worked.thanks you too much. but now i got a problem with DataTypeProvider plugin. its not work with this plugin.

FGCorp:
thanks a lot its worked for me.

PR #906, opened 2018-04-10
chore(vue-grid): add landing page

kvet:


MaximKudriavtsev:
I think that we should change the text color from blue to green.
Seems that current Vue banner is very dark.

PR #909, opened 2018-04-10
chore: fix incorrect row adding

### LINKED ISSUE #904: Issue with React Grid Controlled State Mode

### Current Behaviour
In https://devexpress.github.io/devextreme-reactive/react/grid/demos/featured/controlled-mode/
Starting from a fresh 'no data' table. When we add two rows it works fine. But when we click Add to try adding a third row, the moment we click the Add button, a new row is added with the previous data.

### Expected Behaviour

 clicking the Add button shouldnt add a new row, only when the user clicks save, a new row should be added.

Starting from a fresh 'no data' table. When we add two rows it works fine. But when we click Add to try adding a third row, the moment we click the Add button, a new row is added with the previous data.

### Your Environment
It happens on the website link itself.

devextreme-reactive: xxx
react: xxx
browser: xxx
bootstrap: none
react-bootstrap: none
material-ui: none

kvet:
fixes #904


packages/dx-react-demos/src/demo-sources/grid-editing/disable-column-editing.jsxt
```diff
@@ -28,7 +28,7 @@ export default class Demo extends React.PureComponent {
       ],
       rows: generateRows({
         columnValues: { id: ({ index }) => index, ...defaultColumnValues },
-        length: 8,
+        length: 0,
```
SergeyAlexeev: Rows length is 0. Demo shows 'No data' now.

PR #930, opened 2018-04-12
[ts-generator] Better className and style types

### LINKED ISSUE #922: [Typescript] `style` prop should have `React.CSSProperties` rather than `object`

### Current Behaviour
The type of `style` prop under several Typescript `interface`s are marked as `object`

### Expected Behaviour

They should be marked as `React.CSSProperties` for proper type inference and autocompletion.

### Your Environment

devextreme-reactive: 1.1.2
react: 16.2.0
material-ui: yes

franklixuefei:
fixes #922


franklixuefei:
@gsobolev One other question - 
https://github.com/DevExpress/devextreme-reactive/blob/38308e9342c9da3fb04072143d82f676fe000337/scripts/ts-generator.js#L106
Why do we need a replacement here?

franklixuefei:
@kvet @gsobolev Any thoughts on this? BTW, do you generally accept pull requests from the outside world?

gsobolev:
@franklixuefei,
 
We have discussed this change and are going to merge it. Thank you for helping us make the Grid better.
 
As for your previous question related to the replacement. TypeScript does not allow such a construction as `{ [key: number | string]: any }`. It requires using only a string or a number for the index parameter. That is why this replacement is required.

franklixuefei:
@gsobolev From TypeScript 2.9.1, we do not need to constrain the key name to be only `string`. It now can be `string | number | symbol`. Just as a reminder when you guys decide to support TypeScript 2.9 :-D

gsobolev:
@franklixuefei, Thank you for pointing out to this. We'll take this into account when we support the new TypeScript version.

PR #940, opened 2018-04-16
chore: change last column width in band demo

MaximKudriavtsev:


SergeyAlexeev:
It looks like the last column has a redundant border that should be removed.

MaximKudriavtsev:
No. This border displays the end of a cell and helps users change a cell width.

PR #948, opened 2018-04-17
minColumnWidth isRequired creates warnings

rememberlenny:
Fix the issue by defaulting with a initial value.


kvet:
Hi,

The TableColumnResizing plugin should be used from a theme package. See breaking changes for the [1.0.0-beta.1](https://github.com/DevExpress/devextreme-reactive/blob/master/CHANGELOG.md#100-beta1-2017-11-10).

PR #956, opened 2018-04-18
chore: add the Paging feature

MaximKudriavtsev:


packages/dx-vue-demos/src/demo-sources/grid-paging/none/basic.js
```diff
@@ -0,0 +1,46 @@
...
+import { generateRows } from '../../../demo-data/generator';
+
+export default {
+  data() {
+    return {
+      currentPage: 0,
+      pageSize: 5,
+    };
+  },
+  render() {
```
kvet: Write demos using vue templates

packages/dx-vue-grid-bootstrap4/src/templates/paging-panel/pagination.js
```diff
@@ -0,0 +1,203 @@
+import { firstRowOnPage, lastRowOnPage, calculateStartPage } from '@devexpress/dx-grid-core';
+
+const renderPageButtons = (
```
kvet: Maybe its better to convert this method to a component?

MaximKudriavtsev: No, because Vue can't return a markup array.

kvet: You can use functional component. It allows to return multiple root nodes

PR #965, opened 2018-04-19
chore(vue-grid): add the TableSelection plugin

MaximKudriavtsev:


packages/dx-vue-grid/src/plugins/table-selection.js
```diff
@@ -0,0 +1,146 @@
...
+  isSelectAllTableCell,
+  isDataTableRow,
+} from '@devexpress/dx-grid-core';
+
+export const TableSelection = {
+  name: 'TableSelection',
+  props: {
+    highlightRow: {
+      type: Boolean,
+      default: false,
```
SergeyAlexeev: I think `default: false,` isn't required here.

PR #971, opened 2018-04-20
chore(vue-grid): add the Selection feature

MaximKudriavtsev:


packages/dx-vue-grid-bootstrap4/src/templates/parts/selection-control.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    checked: {
+      type: Boolean,
+    },
+    indeterminate: {
+      type: Boolean,
+    },
+  },
+  updated() {
+    this.$refs['my-ref'].indeterminate = this.indeterminate;
```
kvet: Can we extract this line to a method?

PR #977, opened 2018-04-23
chore(vue-grid): add the EditingState plugin

MaximKudriavtsev:


packages/dx-vue-grid/src/plugins/editing-state.js
```diff
@@ -0,0 +1,176 @@
...
+    createRowChange: {
+      type: Function,
+    },
+    columnEditingEnabled: {
+      type: Boolean,
+      default: true,
+    },
+    columnExtensions: {
+      type: Array,
+      default: undefined,
```
SergeyAlexeev: I'm not sure we should specify `undefined` as default prop value. It's `undefined` by default.

PR #979, opened 2018-04-24
chore(vue-grid): add the TableEditColumn plugin

MaximKudriavtsev:


packages/dx-vue-grid/src/plugins/table-edit-column.test.js
```diff
@@ -0,0 +1,457 @@
...
+              />
+            </PluginHost>
+          );
+        },
+      });
+
+      expect(findCommandWithId(tree, 'delete').at(0).vm.$attrs)
+        .toMatchObject({ text: 'deleteCommand' });
+
+      const { execute } = findCommandWithId(tree, 'delete').at(0).vm.$listeners;
```
SergeyAlexeev: Use `$emit` instread.

PR #1001, opened 2018-04-27
chore(vue-grid): add the Custom Editors demo

MaximKudriavtsev:


packages/dx-vue-demos/src/demo-sources/grid-data-types/none/editors.js
```diff
@@ -0,0 +1,146 @@
...
+  methods: {
+    onChange(event) {
+      this.$attrs.onValueChange(event.target.value);
+    },
+  },
+  template: `
+    <select
+      class="form-control"
+      :value="value"
+      v-on:change="onChange"
```
SergeyAlexeev: Can we use `@change` as we do in other demos? For example [here](https://github.com/DevExpress/devextreme-reactive/pull/1002/files#diff-a7c36ccd3b35c6ae24634341f5626279R31).

PR #1075, opened 2018-05-15
chore(react-chart): min/max applying for scale processing

Krijovnick:


packages/dx-chart-core/src/plugins/integrated-scale-processing/computeds.js
```diff
@@ -6,11 +6,10 @@ const isDefined = item => item !== undefined;
 const getAxesDomains = axes =>
```
alexey-semikozov: Rename this function OR remove it. Now it does nothing

packages/dx-chart-core/src/plugins/integrated-scale-processing/computeds.js
```diff
@@ -55,7 +54,28 @@ const calculateDomain = (series, data, axesDomains, argumentAxisName) =>
     axesDomains,
   );
 
+const zoomingDomains = (axes, calculatedDomains) => axes.reduce(
```
alexey-semikozov: I think "zoomingDomains" is not a good name for this function.

packages/dx-chart-core/src/plugins/integrated-scale-processing/computeds.js
Krijovnick: What about 'collectAxesTypes' name?

packages/dx-chart-core/src/plugins/integrated-scale-processing/computeds.js
Krijovnick: Maybe 'adjustDomains'?

packages/dx-chart-core/src/plugins/integrated-scale-processing/computeds.js
alexey-semikozov: nice 👍

packages/dx-chart-core/src/plugins/integrated-scale-processing/computeds.js
alexey-semikozov: good

PR #1088, opened 2018-05-16
chore(react-chart): calculate d-attribute in 'seriesComponent'

Krijovnick:


packages/dx-react-chart-bootstrap4/src/templates/series/bar.jsx
```diff
@@ -8,10 +7,3 @@ export class Bar extends React.PureComponent {
     );
   }
 }
-
```
alexey-semikozov: Why did you remove Bar.propTypes? Can render bar without x,y, width, height props?

PR #1105, opened 2018-05-21
chore(react-chart): add title component

alexey-semikozov:


packages/dx-react-chart-bootstrap4/src/plugins/title.jsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import { Title as TitleBase } from '@devexpress/dx-react-chart';
+import { Text } from '../templates/title/text';
+
+export class Title extends React.PureComponent {
+  render() {
+    return (
+      <TitleBase
+        titleComponent={Text}
```
MaximKudriavtsev: Maybe rename this property to `textComponent` or `rootComponent` ?

packages/dx-react-demos/src/demo-sources/chart-basic/basic.wb3.jsxt
```diff
@@ -35,6 +36,7 @@ export default class Demo extends React.PureComponent {
           height={height}
         >
 
+          <Title text="Basic chart" />
```
MaximKudriavtsev: Titles should be meaningful. They are dependent on chart data.

packages/dx-react-chart-bootstrap4/src/plugins/title.jsx
alexey-semikozov: textComponent, nice

PR #1116, opened 2018-05-23
chore(react-chart): fix design

alexey-semikozov:


packages/dx-react-chart-material-ui/src/templates/legend/item.jsx
```diff
@@ -7,6 +7,8 @@ import ListItem from '@material-ui/core/ListItem';
 const styles = () => ({
   root: {
     alignItems: 'center',
+    paddingTop: 4,
+    paddingBottom: 4,
```
MaximKudriavtsev: In the Material Design theme, we often write `theme.spacing.unit` instead of numbers. See [example](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-grid-material-ui/src/templates/table-cell.jsx#L10).

packages/dx-chart-core/src/plugins/theme-manager/computeds.js
```diff
@@ -0,0 +1,10 @@
+
+const Palette = ['#2196F3', '#F44336', '#4CAF50', '#FFEB3B', '#E91E63', '#9C27B0'];
```
MaximKudriavtsev: I think this constant should be in theme packages because there are differences.

MaximKudriavtsev:
Is a bug?
image

I think the `group-bar.wb3` demo looks better
image

packages/dx-chart-core/src/plugins/theme-manager/computeds.js
alexey-semikozov: Hi. We have only one palette for MUI & Bootstrap4 packages. MB we rethink it after

PR #1142, opened 2018-05-30
chore(react-chart): make more simple example in getting started guide

Krijovnick:


packages/dx-react-chart/docs/guides/getting-started.md
```diff
@@ -63,10 +63,8 @@ const App = () => (
         { argument: 2, value: 20 },
         { argument: 3, value: 30 }
       ]}
-      width={400}
```
alexey-semikozov: Too early

PR #1154, opened 2018-06-01
chore(react-demos): fix Bootstrap3 Featured Custom demo

gsobolev:


MaximKudriavtsev:
image
Virtual Scrolling Demo is broken

Maybe use default theme for `custom` case?

PR #1162, opened 2018-06-01
chore(react-chart): add legend/title guide

alexey-semikozov:
Guide explains legend/title common use


packages/dx-react-chart/docs/guides/legend_title.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Legend
+
+The Legend feature displays series on the current chart. Legend contains items. Items contain marker and label. Marker display series color. Label text equal series name.
+
+## Title 
+
+Title display helpful text. You must provide text to the title.
+
+## Related Plugins
```
gsobolev: Let's move this section up to make it the first one like it is done in other guides.
Do we really need separate sections for _Legend_ and _Title_? There is no much to say about them and it would be also strange to create separate demos for them. So, I believe we can just extend the common description a little bit and do not create these sections at all.

packages/dx-react-chart/docs/guides/legend_title.md
```diff
@@ -0,0 +1,27 @@
...
+
+- `Legend` - renders a legend
+- `Title` - renders a title
+
+## Basic Setup
+Import the plugins listed above to set up a Chart with the legend and title.
+
+.embedded-demo({ "path": "chart-basic/basic-legend.wb3", "showThemeSelector": true })
+
+## Use custom legend markers
```
SergeyAlexeev: Can we name this section 'Appearance Customization' and show how to customize both the `Title` and `Legend` plugins?

PR #1189, opened 2018-06-14
chore(vue-grid): add the Featured Virtual Scrolling demo

MaximKudriavtsev:


packages/dx-vue-demos/src/theme-sources/none/components/boolean-type-provider.js
```diff
@@ -0,0 +1,37 @@
...
+  `,
+};
+
+const BooleanEditor = {
+  props: ['value'],
+  template: `
+    <select
+      class="form-control"
+      :value="value"
+      @change="e => this.$attrs.onValueChange(e.target.value === 'true')"
```
SergeyAlexeev: Can we use $emit?

PR #1193, opened 2018-06-14
chore(dx-react-chart-docs): update the Material UI link

oliviertassinari:


MaximKudriavtsev:
@oliviertassinari, do you want to update the same code in the `dx-react-grid` package?

oliviertassinari:
What do you mean?

MaximKudriavtsev:
I mean your last commit `Material-UI is the official name`.

oliviertassinari:
Yes, we would need to go through all the documentation and replace all the occurrences. I haven't done it, just replaced the one I have seen in this file.

PR #1229, opened 2018-07-05
chore(react-scheduler): add the Scheduler

MaximKudriavtsev:


packages/dx-react-scheduler/src/plugins/week-view.test.jsx
```diff
@@ -0,0 +1,266 @@
...
+            timePanelLayoutComponent={() => <div className="time-panel" />}
+          />
+        </PluginHost>
+      ));
+
+      expect(tree.find('.time-panel').exists())
+        .toBeTruthy();
+    });
+
+    it('Should render date table', () => {
```
kvet: We write test names in lowercase

packages/dx-scheduler-core/src/utils.js
```diff
@@ -0,0 +1,50 @@
...
+  const leftBound = moment(start.day(day));
+  return [
+    leftBound,
+    moment(leftBound).hour(start.hour()).endOf('day'),
+  ];
+};
+
+const excludedIntervals = (excludedDays, start) => excludedDays
+  .map(day => (day === 0 ? 7 : day))
+  .sort()
```
gsobolev: Please note that `Array.prototype.sort()` default sort order is based on string Unicode codes, so it won't work for numbers correctly. See more details [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort).

packages/dx-react-scheduler-material-ui/src/templates/utils.jsx
```diff
@@ -0,0 +1,7 @@
+import { darken, fade, lighten } from '@material-ui/core/styles/colorManipulator';
+
+export const getBorderColor = theme => (`1px solid ${
+  theme.palette.type === 'light'
+    ? lighten(fade(theme.palette.divider, 1), 0.88)
+    : darken(fade(theme.palette.divider, 1), 0.8)
```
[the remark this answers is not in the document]
MaximKudriavtsev: Material-ui `TableCell` has not border. We use same function `getBorderColor` for create border color for the BandColumns feature.

PR #1284, opened 2018-08-02
chore(grid): remove unnecessary scrollbars in demos

kvet:


MaximKudriavtsev:
Please, remove one column from `Fundamentals` guide and specifies the column width Total to 90 px in `Banded Columns`.

What happens with `redux-integration` demo bs4? Width all of they columns less of the demo width, but scrollbar is available =/

kvet:
I'm afraid that I could not fix the "redux integration" demo. Seems like problems with the border collapsing.

PR #1315, opened 2018-08-14
chore(react): update to new lifecycles

kvet:


packages/dx-grid-core/src/utils/table.test.js
```diff
@@ -38,15 +37,4 @@ describe('table utils', () => {
       expect(getTableTargetColumnIndex(columnGeometries, 1, 260)).toEqual(3);
     });
   });
-
-  describe('#getAnimations', () => {
-    it('should not return animations if columns are the same', () => {
```
alexey-semikozov: Why you delete this test? The getAnimation function has not cover with a test.

kvet: This test is no longer needed. There are no other tests for this function.

packages/dx-react-grid-demos/index.html
```diff
@@ -8,7 +8,7 @@
   <link href="https://fonts.googleapis.com/css?family=Roboto" rel="stylesheet">
 </head>
 <body>
-  <div class="embedded-demo" data-options='{ "router": "hash", "showThemeSelector": true }'>Loading...</div>
+  <div class="embedded-demo" data-options='{ "router": "hash", "showThemeSelector": true, "frame": true }'>Loading...</div>
```
gsobolev: Why are you changing the `frame` parameter here?

kvet: Mistake

PR #1320, opened 2018-08-15
chore(site): add cross reference to grid/chart/core/vue/react

alexey-semikozov:


site/_data/docs/navigation.yml
```diff
@@ -27,8 +27,12 @@
         path: /react/core/docs/reference/template-connector
   - title: Related Docs
     items:
-      - title: Grid
+      - title: React Grid
         path: /react/grid/docs/guides/getting-started
+      - title: React Chart
+        path: /react/chart
```
kvet: Use long paths. This will avoid redirects

site/_data/docs/navigation.yml
```diff
@@ -203,6 +211,14 @@
         path: /react/chart/docs/reference/title
       - title: ValueAxis
         path: /react/chart/docs/reference/value-axis
+  - title: Related Docs
+    items:
+      - title: Core
```
MaximKudriavtsev: We have only `React Core` documentation

PR #1328, opened 2018-08-17
fix(react): fix adding existing plugins

Krijovnick:


kvet:
We should change label to 'fix' and find better name. Is seems that the bug caused the following problem: https://github.com/DevExpress/devextreme-reactive/issues/1329

PR #1390, opened 2018-09-10
chore(react-grid): rename props in fixed columns

kvet:


SergeyAlexeev:
Should we add BC?

kvet:
It is unreleased functionality. So we shouldn't.

packages/dx-react-grid/src/plugins/table-fixed-columns.test.jsx
```diff
@@ -13,7 +13,7 @@ import {
 import { TableFixedColumns } from './table-fixed-columns';
 
 jest.mock('@devexpress/dx-grid-core', () => ({
-  FIXED_COLUMN_BEFORE_SIDE: 'BEFORE',
+  FIXED_COLUMN_LEFT_SIDE: 'BEFORE',
```
SergeyAlexeev: 'BEFORE' -> 'LEFT' ?

PR #1391, opened 2018-09-10
chore(react-grid): merge fixed types and names

kvet:


packages/dx-grid-core/src/plugins/table-row-detail/constants.js
```diff
@@ -1 +1 @@
-export const TABLE_DETAIL_TYPE = 'detail';
+export const TABLE_DETAIL_TYPE = Symbol('detail');
```
MaximKudriavtsev: Why we don't use `.toString()` here?

kvet: We use toString only in the key definition. Type should be symbol.

packages/dx-react-grid-demos/src/demo-sources/grid-fixed-columns/with-selection.jsxt
```diff
@@ -65,9 +63,8 @@ export default class Demo extends React.PureComponent {
           <TableHeaderRow />
           <TableSelection />
           <TableFixedColumns
-            leftColumnTypes={leftColumnTypes}
-            leftColumnNames={leftColumnNames}
-            rightColumnNames={rightColumnNames}
+            leftColumns={leftColumns}
+            rightColumns={rightColumns}
```
MaximKudriavtsev: The sentence `You can also use the leftColumnTypes and rightColumnTypes properties to fix service columns, such as Selection Column or Editing Column.` is keeping in the doc

kvet: Fixed

PR #1393, opened 2018-09-10
feat(react-chart): add legend for pie chart

Krijovnick:


packages/dx-chart-core/src/plugins/series/computeds.js
```diff
@@ -165,3 +165,10 @@ export const seriesData = (series = [], seriesProps) => {
 };
 
 export const checkZeroStart = (fromZero, axisName, pathType) => ({ ...fromZero, [axisName]: fromZero[axisName] || (pathType === 'area' || pathType === 'bar') });
+
+export const getItemsCallback = (pluginName) => {
+  if (pluginName === 'PieSeries') {
```
MaximKudriavtsev: Usually we extract string values to `const` variables

packages/dx-react-chart/src/plugins/area-series.test.jsx
```diff
@@ -14,6 +14,7 @@ jest.mock('@devexpress/dx-chart-core', () => ({
   coordinates: jest.fn(),
   seriesData: jest.fn(),
   checkZeroStart: jest.fn(),
+  getItemsCallback: jest.fn,
```
MaximKudriavtsev: Why we use `jest.fn` instead of `jest.fn()` here?

packages/dx-react-chart/src/utils/series-helper.jsx
```diff
@@ -48,6 +48,7 @@ export const withSeriesPlugin = (
         <Plugin name={pluginName}>
           <Getter name="series" computed={getSeriesDataComputed} />
           <Getter name="startFromZero" computed={startFromZeroByAxes} />
+          <Getter name="items" value={getItemsCallback(pluginName)} />
```
MaximKudriavtsev: I think this is a unclear name

packages/dx-react-chart/src/plugins/area-series.test.jsx
Krijovnick: That's my fault. I'll fix it

packages/dx-chart-core/src/plugins/series/computeds.js
Krijovnick: Ok

packages/dx-chart-core/src/plugins/series/computeds.test.js
```diff
@@ -343,3 +344,9 @@ describe('checkZeroStart', () => {
     expect(fromZero).toEqual({ axis1: false });
   });
 });
+
+describe('get items', () => {
```
MaximKudriavtsev: Usually we use for describing message the following text `#getPieItems`. [Link](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-grid-core/src/plugins/custom-grouping/computeds.test.js#L19)

PR #1395, opened 2018-09-11
chore(react-grid): extract group button in TableHeaderRow

kvet:


packages/dx-react-grid-material-ui/src/templates/table-header-cell/group-button.test.jsx
```diff
@@ -0,0 +1,69 @@
+import * as React from 'react';
+import { createShallow, getClasses } from '@material-ui/core/test-utils';
+import { GroupButton } from './group-button';
+
+const defaultProps = {
+  onGroup: jest.fn(),
+  disabled: false,
```
MaximKudriavtsev: The `disable` property is not a required prop. The same in other tests

packages/dx-react-grid/docs/reference/table-header-row.md
```diff
@@ -90,8 +91,17 @@ Describes properties used to render a cell content.
...
 ------|------|------------
 column | [Column](grid.md#column) | A column object associated with the cell content.
-align? | string | The content alignment.
-children? | ReactNode | The content's children.
+align | string | The content alignment.
+children | ReactNode | The content's children.
+
+### TableHeaderRow.GroupButtonProps
+
+Describes propertiesused to render a group button.
```
SergeyAlexeev: 'propertiesused' => 'properties used'

SergeyAlexeev:
What about a demo like [that](https://devexpress.github.io/devextreme-reactive/@next/react/grid/docs/guides/sorting#custom-sort-label)?

kvet:
I don't think that this component customisation is required. It will not be often customized

PR #1400, opened 2018-09-13
refactor(react-chart): remove theme related copypaste

DmitryBogomolov:


kvet:
Hm. I don't like the change. We reduced codebase only by a half but increased complexity dramatically.

DmitryBogomolov:
What are the points where complexity has been increased?

packages/dx-react-chart-bootstrap4/src/templates/axis/tick.jsx
```diff
@@ -1,33 +1,4 @@
...
-  x2: PropTypes.number.isRequired,
-  y1: PropTypes.number.isRequired,
-  y2: PropTypes.number.isRequired,
-  className: PropTypes.string,
-};
-
-Tick.defaultProps = {
-  className: undefined,
-};
+export const Tick = withClassName('dx-c-bs4-stroke-current-color dx-c-bs4-crisp-edges dx-c-bs4-axis-opacity')(Axis.Tick);
```
SergeyAlexeev: Can we support `array` syntax here to avoid too long strings? I mean:

```
withClassName([
'a',
'b',
'c',
...
])
```

packages/dx-react-chart-material-ui/src/templates/legend/label.test.jsx
```diff
@@ -1,22 +1,22 @@
...
-import { createShallow, getClasses } from '@material-ui/core/test-utils';
+import { createMount, getClasses } from '@material-ui/core/test-utils';
 import ListItemText from '@material-ui/core/ListItemText';
 import { Label } from './label';
 
 const defaultProps = { text: 'a' };
 
 describe('Label', () => {
-  const shallow = createShallow({ dive: true });
+  const mount = createMount({ dive: true });
```
SergeyAlexeev: I'm not sure `{ dive: true }` is a required option for `mount`.

packages/dx-react-chart/src/plugins/axis.jsx
```diff
@@ -136,7 +143,7 @@ export class Axis extends React.Component {
                       position: 'absolute', left: 0, top: 0, overflow: 'visible',
                     }}
                   >
-                    <Root
+                    <RootComponent
```
SergeyAlexeev: Usually, we prefer don't use *Component postfix for components.

DmitryBogomolov: `Root` is already occupied in the outer scope.

packages/dx-react-chart-material-ui/src/templates/legend/label.test.jsx
DmitryBogomolov: I'll check it.

packages/dx-react-chart-bootstrap4/src/templates/axis/tick.jsx
DmitryBogomolov: Yes, we can. Since there are cases when single string is provided, it seems that it would be better to support arbitrary amount of arguments (like it does underlying [classnames](https://www.npmjs.com/package/classnames)).

packages/dx-react-chart-material-ui/src/templates/legend/label.test.jsx
DmitryBogomolov: Yes, there is no `dive` option for `createMount`.
`createShallow({ dive: true })` cannot be used as it gives [one](https://material-ui.com/guides/testing/#api) level of depth.
Redundant option has been removed.

packages/dx-react-chart-bootstrap4/src/templates/axis/tick.jsx
DmitryBogomolov: Updated.

packages/dx-react-chart/src/plugins/legend.jsx
```diff
@@ -1,17 +1,18 @@
...
   Plugin,
   TemplateConnector,
   Template,
   TemplatePlaceholder,
 } from '@devexpress/dx-react-core';
+import { Marker } from '../templates/legend/marker';
+import { withComponents } from '../utils';
 
-export class Legend extends React.PureComponent {
+export class RawLegend extends React.PureComponent {
```
SergeyAlexeev: What for do we export `RawLegend`? The same about `RawGrid`.

DmitryBogomolov: Fixed (`RawGrid` had no `export`).

packages/dx-react-chart/src/plugins/axis.jsx
```diff
@@ -69,10 +74,10 @@ export class Axis extends React.Component {
       name,
       indentFromAxis,
       isArgumentAxis,
-      rootComponent: Root,
-      tickComponent: Tick,
-      labelComponent: Label,
-      lineComponent: Line,
+      rootComponent: RootComponent,
```
kvet: Why do you use *Component in names?

DmitryBogomolov: `Root` is occupied in the outer scope (`import` section).

packages/dx-react-chart/src/chart.jsx
```diff
@@ -52,16 +54,22 @@ export class Chart extends React.PureComponent {
...
 };
 
-Chart.defaultProps = {
+RawChart.defaultProps = {
   height: 500,
   width: undefined,
   children: null,
 };
+
+RawChart.components = {
```
kvet: Just an idea. Can we use Symbol here?

DmitryBogomolov: You mean?
```javascript
const name = Symbol('components');
RawChart[name] = ...
```

PR #1405, opened 2018-09-17
fix(react): handle event changes correctly

kvet:


packages/dx-react-core/src/utils/state-helper.test-utils.jsx
```diff
@@ -207,6 +207,40 @@ export const testStatePluginField = ({
         expect(change)
           .toHaveBeenCalledTimes(1);
       });
+
+      it(`should correctly work with when ${eventPropertyName} change event changes`, () => {
```
SergeyAlexeev: It looks like there is a typo in a test description. ...work `with` or `when`?

PR #1408, opened 2018-09-18
feat(react-chart): add palette plugin

Krijovnick:


packages/dx-react-chart-demos/src/demo-sources/chart-basic/bootstrap4/theme.jsx
```diff
@@ -0,0 +1,93 @@
...
+  schemePastel1,
+  schemePastel2,
+  schemeSet1,
+  schemeSet2,
+  schemeSet3,
+} from 'd3-scale-chromatic';
+
+import { Scale, Theme } from '@devexpress/dx-react-chart';
+
+const materialUI = ['#2196F3', '#F44336', '#4CAF50', '#FFEB3B', '#E91E63', '#9C27B0'];
```
MaximKudriavtsev: I think that we should use `MaterialDesignColors` instead `MaterialUI` because material-ui is only react library.
Add more colors to that palette from official site https://material.io/design//color/the-color-system.html#tools-for-picking-colors

MaximKudriavtsev:
Can we implement this feature without add `Theme` plugin to markup? It looks excess because the Chart should have colors always.

packages/dx-react-chart-demos/src/demo-sources/chart-basic/bootstrap4/theme.jsx
Krijovnick: I'll change the name to `MaterialDesignColors`. 
And why are we need to add more color to that palette?

MaximKudriavtsev: Because colors are repeated and it looks bad.

Krijovnick:
@MaximKudriavtsev I think this was the point because we have plugin structure.

packages/dx-react-chart-demos/src/demo-sources/chart-basic/bootstrap4/theme.jsx
Krijovnick: But other schemes has not many colors too.

MaximKudriavtsev:
@Krijovnick but this plugin is required. I should always add this plugin to my markup, I think this is problem.

PR #1409, opened 2018-09-19
Remove theme related copypaste in grid and scheduler

DmitryBogomolov:


packages/dx-react-grid/src/plugins/table-column-reordering.jsx
```diff
@@ -25,7 +27,7 @@ const tableHeaderRowsComputed = (
   { tableHeaderRows },
 ) => tableHeaderRowsWithReordering(tableHeaderRows);
 
-export class TableColumnReordering extends React.PureComponent {
+class RawTableColumnReordering extends React.PureComponent {
```
MaximKudriavtsev: You create `GroupingPanelRaw` and `RawTableColumnReordering` names. Can we keep one style for naming `Raw` components. I think that postfix is better.

DmitryBogomolov: Fixed.

PR #1419, opened 2018-09-21
chore(react-scheduler): add the EditingState plugin

MaximKudriavtsev:


packages/dx-scheduler-core/src/plugins/editing-state/computeds.js
```diff
@@ -0,0 +1,14 @@
+export const addedAppointmentById = addedAppointment => addedAppointment;
+
+export const changedAppointmentById = (changes, appointmentId) => { // !
```
SergeyAlexeev: Is it a required comment?

MaximKudriavtsev: No

packages/dx-scheduler-core/src/plugins/editing-state/reducers.js
```diff
@@ -0,0 +1,23 @@
...
+
+export const startEditAppointment = (prevEditingAppointmentId, { appointmentId }) => appointmentId;
+
+export const stopEditAppointment = () => null;
+
+export const changeAppointment = (
+  addedAppointment, { change },
+) => ({ ...addedAppointment, ...change });
+
+export const cancelChanges = () => ({});
```
SergeyAlexeev: Why do we sometimes return an empty object and sometimes null?

MaximKudriavtsev: Because sometimes default value for an object's property is `{}` and sometimes default value for a number's property is `null`.

PR #1435, opened 2018-09-26
fix(react-chart): bind default domain names for argument and value axes

DmitryBogomolov:
BREAKING CHANGES:
Previously, there was a single `Grid` plugin. Now there are `ArgumentGrid` and `ValueGrid` plugins. 
`ArgumentGrid` is bound to an argument domain.
`ValueGrid` is bound to a value domain specified by the name property or to a default domain.


Krijovnick:
Change PR name from 'fix(chart)' to 'fix(react-chart)' pls.

PR #1468, opened 2018-10-05
chore(react-scheduler): add Scheduler editing logic

MaximKudriavtsev:


packages/dx-react-scheduler-material-ui/src/templates/appointment-form/date-editor.jsx
```diff
@@ -16,36 +17,43 @@ const DateEditorBase = ({
...
     }}
+    onChange={({ target }) => onValueChange(target.value)}
     {...restProps}
   />
 );
 
 DateEditorBase.propTypes = {
   classes: PropTypes.object.isRequired,
-  value: PropTypes.string,
+  value: PropTypes.oneOfType([
```
SergeyAlexeev: Just a thought. Should we allow using numbers here? It's useful for the [Unix time](https://en.wikipedia.org/wiki/Unix_time) format.

packages/dx-react-scheduler/src/plugins/appointment-form.jsx
```diff
@@ -21,13 +33,19 @@ const defaultMessages = {
...
+];
+
 export class AppointmentForm extends React.PureComponent {
   constructor(props) {
     super(props);
 
     this.state = {
       visible: props.visible,
-      appointment: props.appointment,
+      appointment: {} || props.appointment,
```
SergeyAlexeev: Hm, this line looks strange. `appointment` is always `{}`;

MaximKudriavtsev: A default value for appointment is `undefined` because this case correctly work with [this](https://github.com/DevExpress/devextreme-reactive/pull/1468/files#diff-eb36d6af4dfa29e1519a9a3731f05b67R77)

PR #1481, opened 2018-10-10
chore(react-chart): fix chart doc reference on the site

Krijovnick:


site/_data/docs/navigation.yml
```diff
@@ -215,8 +213,6 @@
         path: /react/chart/docs/reference/bar-series
       - title: Chart
         path: /react/chart/docs/reference/chart
-      - title: Grid
-        path: /react/chart/docs/reference/grid
```
DmitryBogomolov: Shouldn't `ArgumentGrid` and `ValueGrid` entries be added?

PR #1483, opened 2018-10-11
refactor(react-chart): add function that gets series points

DmitryBogomolov:


Krijovnick:
Change title of the PR, pls: chart -> react-chart

Krijovnick:
Demo with pie chart and palettes is not work. Palette is not apply on pie

img

PR #1498, opened 2018-10-17
 chore(react-chart): add chart text component

Krijovnick:


packages/dx-react-chart/src/chart.jsx
```diff
@@ -24,6 +25,7 @@ class RawChart extends React.PureComponent {
       height,
       children,
       rootComponent,
+      labelComponent,
```
DmitryBogomolov: Where is it used in markup?

PR #1499, opened 2018-10-17
chore(react-scheduler): apply many click handlers to the Appointment template

MaximKudriavtsev:


packages/dx-core/src/index.js
```diff
@@ -4,3 +4,4 @@ export { memoize } from './memoize';
 export { shallowEqual, argumentsShallowEqual } from './shallow-equal';
 export * from './easings';
 export { getMessagesFormatter } from './messages-formatter';
+export { createHandlers } from './utils';
```
SergeyAlexeev: createEventHandlers

packages/dx-core/src/utils.js
```diff
@@ -17,3 +19,24 @@ export const insertPlugin = (array, newItem) => {
   result.splice(targetIndex, alreadyExists ? 1 : 0, newItem);
   return result;
 };
+
+export const createHandlers = (clickEvent, dblClickEvent) => {
```
SergeyAlexeev: createHandlers -> createEventHandlers

packages/dx-core/src/utils.js
```diff
@@ -17,3 +19,24 @@ export const insertPlugin = (array, newItem) => {
...
+  let timeoutId;
+  const onClick = clickEvent ? (payload) => {
```
SergeyAlexeev: Usually, we use the `e` or `evt` name as an event handler argument.

PR #1524, opened 2018-10-23
chore(react-chart): add hover and selection states for series and points

DmitryBogomolov:
Adds `state` field to series and points objects from *series* getter.
State can be *hovered* or *selected* or undefined.


packages/dx-chart-core/src/utils/series.js
```diff
@@ -80,3 +80,39 @@ export const createPieHitTester = coordinates => ([px, py]) => {
...
+    if (point >= 0) {
+      result[series].points[point] = true;
+    } else {
+      result[series].self = true;
+    }
+  });
+  return result;
+};
+
+export const changeSeriesState = (seriesList, targets, state) => {
```
Krijovnick: There is no test on this function

PR #1527, opened 2018-10-23
chore(react-scheduler): use transparent color for appointment borders

MaximKudriavtsev:


packages/dx-react-scheduler-material-ui/src/templates/appointment/appointment.jsx
```diff
@@ -6,10 +6,11 @@ import { withStyles } from '@material-ui/core/styles';
...
   appointment: {
     overflow: 'hidden',
-    backgroundColor: palette.primary[300],
     boxSizing: 'border-box',
-    borderRight: `1px solid ${palette.background.paper}`,
-    borderBottom: `1px solid ${palette.background.paper}`,
+    borderRight: '1px solid transparent',
+    borderBottom: '1px solid transparent',
+    backgroundClip: 'padding-box',
+    backgroundColor: palette.primary[300],
```
SergeyAlexeev: We had some problems with such colors, didn't we?

MaximKudriavtsev: Yes

PR #1532, opened 2018-10-25
fix(grid): fix banded columns layout in virtual table

### LINKED ISSUE #1389: Banded columns and virtual table bug
I use VirtualTable and BandedColumns together, column header cells behave strangely when horizontal scrolling. I think colspan is calculated incorrectly for banded columns. When the first column from group is removed from DOM, its banded header cell it is also removed

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] Vue Grid

ushkal:
fixes #1389

### DIFF
--- a/packages/dx-grid-core/src/plugins/table-band-header/helpers.js
+++ b/packages/dx-grid-core/src/plugins/table-band-header/helpers.js
@@ -10,7 +10,7 @@ export const isBandedOrHeaderRow = tableRow => isBandedTableRow(tableRow)
   || tableRow.type === TABLE_HEADING_TYPE;
 export const isNoDataColumn = columnType => columnType !== TABLE_DATA_TYPE;
 
-export const getColumnMeta = (
+export const getColumnBandInfo = (
   columnName, bands, tableRowLevel,
   level = 0, title = null, result = null,
 ) => bands.reduce((acc, column) => {
@@ -20,7 +20,7 @@ export const getColumnMeta = (
     return acc;
   }
   if (column.children !== undefined) {
-    return getColumnMeta(
+    return getColumnBandInfo(
       columnName,
       column.children,
       tableRowLevel,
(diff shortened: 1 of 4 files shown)


packages/dx-grid-core/src/plugins/table-band-header/helpers.js
```diff
@@ -31,87 +32,158 @@ export const getColumnMeta = (
...
+  return (prevColumnBand.title === currentColumnBand.title
+    && !currentTableColumn.rightColumnIndex
+    && (!isPrevColumnFixed || (isPrevColumnFixed && isCurrentColumnFixed)));
+};
+
+const columnRenderedInOtherRow = ({ currentColumnBand, currentRowLevel }) => (
+  currentColumnBand.level < currentRowLevel
+);
+
+const columnBelongsToCurrentRow = ({ currentColumnBand, currentRowLevel }) => (
```
MaximKudriavtsev: Can we rename this functions similarly? I mean `columnRenderedInOtherRow` and (`columnBelongsToCurrentRow` => `columnRenderedInCurrentRow`)

ushkal: these functions contain slightly different logic: 
`columnRenderedInOtherRow` means that a current header cell has _already_ been rendered in some upper row (and we replace it with empty cell), 
`columnBelongsToCurrentRow` returns true for a cell that has not been yet rendered.
We can consider names like `shouldRenderColumnInCurrentRow`, `isColumnInCurrentRow` or something similar

MaximKudriavtsev: I prefer `isColumnInCurrentRow` or `isCurrentRowColumn`

SergeyAlexeev: `shouldRenderColumnInCurrentRow` and `isColumnInCurrentRow`

+1

PR #1533, opened 2018-10-25
fix(react-grid): correct order of fixed right columns

ushkal:


SergeyAlexeev:
Let's use the `fix(react-grid):` prefix in the PR name.

PR #1538, opened 2018-10-29
chore(react-scheduler): merge appointment get functions

MaximKudriavtsev:


packages/dx-react-scheduler-demos/src/demo-sources/scheduler-featured-remote-data/material-ui/demo.jsx
```diff
@@ -108,9 +111,7 @@ export default class Demo extends React.PureComponent {
       <Paper>
         <Scheduler
           data={data}
-          getAppointmentStartDate={getAppointmentStartDate}
-          getAppointmentEndDate={getAppointmentEndDate}
-          getAppointmentTitle={getAppointmentTitle}
+          appointmentMapping={appointmentMapping}
```
SergeyAlexeev: `appointmentMapping` -> `getAppointmentData`

Discussed with @viterobk

MaximKudriavtsev: Can we use the `getAppointmentData` name for the Getter and `appointmentMapping` for a property?

I think we should not use a `appointmentData` property name, because a `appointmentData` property has all custom fields

for example
```js
{
  STARTDATE: ...,
  _EndDate: ...,
  ...
}
``` 

but in this case, the `getAppointmentData` function will always return a specific object with defined fields

```js
{
  startDate: ...,
  endDate: ...,
  id: ...,
  title: ...
}
```

PR #1546, opened 2018-10-31
chore(scheduler-core): fix the adjustAppointments function with byDay

MaximKudriavtsev:
Before

image

After

image


packages/dx-scheduler-core/src/utils.js
```diff
@@ -90,6 +90,10 @@ export const findOverlappedAppointments = (sortedAppointments, byDay = false) =>
   return groups;
 };
 
+const isMidnight = date => date.isSame(date.clone().startOf('day'));
+const byDayConditional = (maxBoundary, startDate) => ((maxBoundary.isBefore(startDate, 'day'))
```
SergeyAlexeev: `byDayConditional` looks a bit strange. Can we use something like `maxBoundaryPredicate`?

MaximKudriavtsev: Renamed

PR #1566, opened 2018-11-02
chore(react-scheduler): refactor get/set functions

MaximKudriavtsev:


packages/dx-react-scheduler/src/plugins/appointment-tooltip.test.jsx
```diff
@@ -26,6 +21,9 @@ describe('AppointmentTooltip', () => {
     headerComponent: () => null,
     contentComponent: () => null,
     commandButtonComponent: () => null,
+    appointmentMeta: {
```
SergeyAlexeev: appointmentMate should follow the following shape:
`{ data: {}, target: {} }`

BTW, I found the `getAppointmentId: jest.fn(),` mock in this file. Seems like it should be removed.

PR #1567, opened 2018-11-02
feat(react-grid): provide a way to customize a summary item

### LINKED ISSUE #1525: Table Summary Row - display text
In the various aggregations in the summary row plugin the default display text is of the form "Sum:  1,234" or "Count: 23".

In situations where i am using it, it is clear from the context that they are all Sums, so seeing  "Sum: xxxx" in each column doesn't look great.

I know that I can use the Localization messages [https://devexpress.github.io/devextreme-reactive/react/grid/docs/reference/table-summary-row/#localization-messages](url)  to remove the "Sum" part but it still leaves me with displays of the form ":  xxxx".

From [https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-grid/src/plugins/table-summary-row.jsx](url) I can see that the ":&nbsp;&nbsp" is hard-coded.

`<React.Fragment>
        {columnSummaries.map((summary) => {
           if (summary.value === null
            || formatlessSummaryTypes.includes(summary.type)
            || defaultTypelessSummaries.includes(summary.type)) {
            return (
              <Item
                key={summary.type}
              >
                {getMessage(summary.type)}
                :&nbsp;&nbsp;
                {String(summary.value)}
              </Item>
            );
          }`

Would it be possible to make that piece of the string customizable too? 

Thanks!

ushkal:
closes #1525


SergeyAlexeev:
Use the `feat(react-grid):` prefix in PR name.

packages/dx-react-grid-bootstrap4/src/templates/table-summary-item.test.jsx
```diff
@@ -27,10 +34,28 @@ describe('TableSummaryItem', () => {
 
   it('should pass rest props to the root element', () => {
     const tree = shallow((
-      <TableSummaryItem className="custom-class" />
+      <TableSummaryItem
+        {...defaultProps}
+        className="custom-class"
```
SergeyAlexeev: I think, in this test we should check both classes: `dx-g-bs4-table-summary-item` and `custom-class`. Also, we can rename it to `should pass the className prop to the root element`.

packages/dx-react-grid-bootstrap4/src/templates/table-summary-item.test.jsx
```diff
@@ -17,6 +23,7 @@ describe('TableSummaryItem', () => {
   it('should pass style to the root element', () => {
```
SergeyAlexeev: There isn't any code that merges styles for a summary item. Well, we can have a more common test here. Something like `should pass rest props to the root element`.

PR #1577, opened 2018-11-06
chore(react-chart): update series reference files

DmitryBogomolov:


packages/dx-react-chart/docs/reference/scatter-series.md
```diff
@@ -44,21 +44,10 @@ y | number | The point's y coordinate.
 d | string | The point's [path](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d).
 value | number | The point's value.
 color | string | A series color.
-style | object | Series styles.
```
Krijovnick: In our [demos](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-chart-demos/src/demo-sources/chart-basic/point-customization.jsxt#L97) we are using ScatterSeries.Path and pointComponent.

DmitryBogomolov: Reverted.

PR #1594, opened 2018-11-08
fix(react-scheduler): add cursor pointer to appointment template

MaximKudriavtsev:


packages/dx-react-scheduler-material-ui/src/templates/appointment/appointment.test.jsx
```diff
@@ -70,5 +72,35 @@ describe('Appointment', () => {
...
+          <div />
+        </Appointment>
+      ));
+
+      expect(tree.is(`.${classes.clickableAppointment}`))
+        .toBeTruthy();
+    });
+
+    it('should apply clickable class if onDoubleClick event exists', () => {
+      const clickMock = jest.fn();
```
SergeyAlexeev: The same about mock function.

packages/dx-react-scheduler-material-ui/src/templates/appointment/appointment.test.jsx
```diff
@@ -70,5 +72,35 @@ describe('Appointment', () => {
       expect(clickMock.mock.calls[0][0])
         .toEqual({ target: 'target', data: { text: 'a' } });
     });
+
+    it('should apply clickable class if onClick event exists', () => {
+      const clickMock = jest.fn();
```
SergeyAlexeev: A mock function isn't required here. Usually `jest.fn()` is useful if we want to check function arguments or call count. 
`onClick={() => {}}` is enough.

PR #1595, opened 2018-11-08
chore(react-grid): note about getRowId function in the DetailRow feature guide

MaximKudriavtsev:


packages/dx-react-grid-demos/src/demo-sources/grid-detail-row/detail-row-controlled.jsxt
```diff
@@ -32,7 +37,10 @@ export default class Demo extends React.PureComponent {
         { name: 'city', title: 'City' },
         { name: 'car', title: 'Car' },
       ],
-      rows: generateRows({ length: 8 }),
+      rows: generateRows({
```
ushkal: We can extract a function (maybe 'generateUniqueRows') that will return rows with ids

PR #1596, opened 2018-11-08
feat(react-grid): provide a way to customize a group cell

### LINKED ISSUE #793: Hide column title from TableGroupRow.Cell
Hello!

Is there any way to hide the column name in custom grouping?

I would like to achieve the result of the first group in the image below:
screen shot 2018-03-01 at 13 51 17

I tried to nullify the column title and name properties (see code) but the colon wasn't removed.
```
<TableGroupRow.Cell
    {...props}
    column ={{title: null, name: null}}
>
    <img src="images/profile.png" /> {props.row.value}
</TableGroupRow.Cell>
```

The result was like this:
screen shot 2018-03-01 at 13 57 24

Any way to resolve this?

ushkal:
closes #793


packages/dx-react-grid-bootstrap3/src/templates/table-group-cell/cell.test.jsx
```diff
@@ -0,0 +1,77 @@
+import * as React from 'react';
+import { shallow, mount } from 'enzyme';
+import { setupConsole } from '@devexpress/dx-testing';
+import { Cell } from './cell';
+
+describe('TableCell', () => {
```
MaximKudriavtsev: I think we should add one more describe to all tests with a 'TableGroupCell' message.

> TableGroupCell
>> Cell
>> Context
>> Icon

packages/dx-react-grid/docs/reference/table-group-row.md
```diff
@@ -71,6 +73,24 @@ Field | Type | Description
...
+column | [Column](grid.md#column) | The column associated with the group.
+children? | ReactNode | A React node to be rendered within the cell's content.
+
+### TableGroupRow.IconProps
+
+Describes properties passed to a component that renders a group expand icon.
+
+Field | Type | Description
+------|------|------------
+expanded | boolean | Specifies whether the row is expanded.
```
MaximKudriavtsev: This is not required property

packages/dx-react-grid-material-ui/src/templates/table-group-cell/content.jsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+
+export const Content = ({
+  column, row, className, children,
+}) => (
+  <span className={className}>
```
MaximKudriavtsev: We should configure classNames for this component in one file. And we should combine our `className` and customer's `className`.

In your case, if a customer applies className to this component only one `className` (customer or our) would be applied.

You can see this behaviour, if you will use `contentComponent`
```jsx

const Content = props => (
  <TableGroupRow.Content {...props} className="custom-class-name" />
);
```

Same with other components.

MaximKudriavtsev:
I think we should update our customization demo. For example
image

packages/dx-react-grid/docs/reference/table-group-row.md
```diff
@@ -34,6 +34,8 @@ showColumnsWhenGrouped? | boolean | false | A Boolean value that specifies wheth
 columnExtensions? | Array&lt;[TableGroupRow.ColumnExtension](#tablegrouprowcolumnextension)&gt; | | Additional column properties that the plugin can handle.
 cellComponent | ComponentType&lt;[TableGroupRow.CellProps](#tablegrouprowcellprops)&gt; | | A component that renders a group cell.
 rowComponent | ComponentType&lt;[TableGroupRow.RowProps](#tablegrouprowrowprops)&gt; | | A component that renders a group row.
+contentComponent | ComponentType&lt;[TableGroupRow.ContentProps](#tablegrouprowcontentprops)&gt; | | A component that renders a group cell's content.
+iconComponent | ComponentType&lt;[TableGroupRow.IconProps](#tablegrouprowiconprops)&gt; | | A component that renders a group expand icon.
```
SergeyAlexeev: The `contentComponent` and `iconComponent` aren't pointed in the `Plugin Components` section.

packages/dx-react-grid/docs/reference/table-group-row.md
ushkal: Can't agree here: the only purpose of the Icon component is to display a state of this property

MaximKudriavtsev: In the `propTypes` this property is not required

packages/dx-react-grid-material-ui/src/templates/table-group-cell/icon.jsx
```diff
@@ -0,0 +1,45 @@
...
+      expanded
+        ? <ExpandMore />
+        : <ChevronRight />
+    }
+  </IconButton>
+);
+
+IconBase.propTypes = {
+  expanded: PropTypes.bool.isRequired,
+  className: PropTypes.string,
```
MaximKudriavtsev: We often add a property `classes` to `propTypes`. `classes: PropTypes.object.isRequired,`

PR #1611, opened 2018-11-14
feat(react-chart): add tooltip plugin

Krijovnick:


packages/dx-react-chart/src/templates/tooltip/target.jsx
```diff
@@ -0,0 +1,26 @@
...
+      d, x, y, componentRef,
+    } = this.props;
+    return (
+      <path transform={`translate(${x} ${y})`} fill="none" ref={componentRef} d={d} />
+    );
+  }
+}
+
+Target.propTypes = {
+  x: PropTypes.number,
```
DmitryBogomolov: Why are `x`, `y` and `d` not required?

packages/dx-react-chart/src/plugins/tooltip.jsx
```diff
@@ -0,0 +1,111 @@
...
+        </Template>
+      </Plugin>
+    );
+  }
+}
+
+RawTooltip.propTypes = {
+  defaultTargetItem: PropTypes.shape({
+    series: PropTypes.string.isRequired,
+    point: PropTypes.number,
```
DmitryBogomolov: Since tooltip is defined for points only this field is also required.

packages/dx-react-chart/src/plugins/tooltip.jsx
```diff
@@ -0,0 +1,111 @@
...
+    ];
+  }
+
+  handlePointerMove({ targets }) {
+    const { onTargetItemChange } = this.props;
+    const { target: currentTarget } = this.state;
+    const target = processPointerMove(targets, currentTarget, onTargetItemChange);
+    if (target !== undefined) {
+      if (target && target.point === undefined) {
+        this.setState({
```
DmitryBogomolov: Since calculated state depends on current state callback version of `setState` should be used.

packages/dx-react-chart/src/plugins/tooltip.jsx
```diff
@@ -0,0 +1,111 @@
...
+  withComponents,
+} from '@devexpress/dx-react-core';
+import { getParameters, processPointerMove } from '@devexpress/dx-chart-core';
+import { Target } from '../templates/tooltip/target';
+
+class RawTooltip extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      target: props.targetItem || props.defaultTargetItem,
```
DmitryBogomolov: For *controlled/uncontrolled* approach `getDerivedStateFromProps` should also be defined.

packages/dx-react-chart/src/templates/tooltip/target.jsx
Krijovnick: If there is no target that parameters are absent

packages/dx-react-chart/src/plugins/tooltip.jsx
Krijovnick: done

packages/dx-react-chart/src/templates/tooltip/target.jsx
DmitryBogomolov: If there is no target should `Target` component be rendered at all?

packages/dx-chart-core/src/plugins/tooltip/computeds.js
```diff
@@ -0,0 +1,8 @@
+export const getParameters = (series, target) => {
+  if (target) {
+    const currentSeries = series.find(({ name }) => target.series === name);
+    const item = currentSeries.points[target.point];
```
DmitryBogomolov: `point` is index within `data` array (particular series can have less points then items in `data` array)

Use `points.find(point => point.index === index)` or `d3.bisect`.

packages/dx-chart-core/src/plugins/series/computeds.js
```diff
@@ -159,6 +192,7 @@ const scalePoints = (series, scales) => {
   });
   return {
     ...rest,
+    getTargetElement: getPointTransformer.getTargetElement,
```
DmitryBogomolov: Add *TODO* so we don't forget to remove it when it is possible.

packages/dx-react-chart-bootstrap4/src/templates/tooltip/overlay.jsx
```diff
@@ -0,0 +1,34 @@
...
+        {children}
+      </PopoverBody>
+    </Popover>
+  ) : null
+);
+
+Overlay.propTypes = {
+  children: PropTypes.node.isRequired,
+  visible: PropTypes.bool,
+  target: PropTypes.oneOfType([
```
DmitryBogomolov: It should be `PropTypes.func.isRequired`.

packages/dx-react-chart-bootstrap4/src/templates/tooltip/overlay.jsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import { Popover, PopoverBody } from 'reactstrap';
+
+export const Overlay = ({
+  visible, children, target, ...restProps
+}) => (
+  target ? (
```
DmitryBogomolov: Isn't it always defined?

packages/dx-react-chart/src/plugins/tooltip.jsx
```diff
@@ -0,0 +1,113 @@
...
+                  <TargetComponent
+                    {...element}
+                    componentRef={(ref) => { this.targetElement = ref; }}
+                  />
+                  <OverlayComponent
+                    key={text}
+                    target={this.getTargetElement}
+                    visible={!!target}
+                  >
+                    <ContentComponent text={text} targetItem={target} />
```
DmitryBogomolov: `text` is quite simply built from `target` - let's pass only `targetItem` to `Content` component and build text inside that component.

Krijovnick: In that case we also should pass value from data.

packages/dx-react-chart/src/plugins/tooltip.jsx
Krijovnick: done

PR #1614, opened 2018-11-15
refactor(react-core): use new React Context API

MaximKudriavtsev:


packages/dx-react-core/src/plugin-based/action.jsx
```diff
@@ -50,12 +52,17 @@ export class Action extends React.PureComponent {
...
   name: PropTypes.string.isRequired,
   action: PropTypes.func.isRequired,
-};
-
-Action.contextTypes = {
   [PLUGIN_HOST_CONTEXT]: PropTypes.object.isRequired,
   [POSITION_CONTEXT]: PropTypes.func.isRequired,
 };
+
+export const Action = withPropsFromContext({
```
DmitryBogomolov: This function is called 4 times with same arguments and 1 time with a different second argument.
Its name implies that component is connected to a single context.

How about changing it so that it actually connects component to a single context?
```javascript
const TargetWithContext = withContext(SomeContext, name)(Target);
```

And then introducing `withHostAndPosition` function?
```javascript
const Action = withHostAndPosition(ActionBase);  // Same for Getter, Plugin, Template
```
```javascript
const TemplatePlaceholder = withHost(
  withContext(TemplateHostContext, TEMPLATE_HOST_CONTEXT)(
    TemplatePlaceholderBase
  )
);
```

MaximKudriavtsev: Done

packages/dx-react-core/src/utils/with-props-from-context.test.jsx
```diff
@@ -0,0 +1,73 @@
...
+  );
+  describe('#withHostAndPosition', () => {
+    const expectedContexts = {
+      [PLUGIN_HOST_CONTEXT]: 'pluginHostContext',
+      [POSITION_CONTEXT]: 'positionContext',
+    };
+    it('should provide props from contexts', () => {
+      const TestBase = () => <div />;
+      const Test = withHostAndPosition(TestBase);
+      const tree = mount((<ContextProvider><Test /></ContextProvider>));
```
SergeyAlexeev: You can use single parenthesis in this case.

PR #1619, opened 2018-11-16
fix(react-grid): fix header layout when middle band column is fixed

ushkal:


packages/dx-react-grid/src/plugins/table-filter-row.jsx
```diff
@@ -62,13 +63,17 @@ export class TableFilterRow extends React.PureComponent {
...
+      { tableHeaderRows, tableColumns, tableHeaderColumnChains },
+    ) => tableHeaderColumnChainsWithFilter(tableHeaderColumnChains, tableHeaderRows, tableColumns);
 
     return (
       <Plugin
         name="TableFilterRow"
         dependencies={pluginDependencies}
       >
         <Getter name="tableHeaderRows" computed={tableHeaderRowsComputed} />
+        <Getter name="tableHeaderColumnChains" computed={tableHeaderColumnChainsComputed} />
```
MaximKudriavtsev: Is this getter necessary here?

packages/dx-react-grid/src/plugins/grid-core.jsx
```diff
@@ -21,6 +23,7 @@ export class GridCore extends React.PureComponent {
         <Getter name="getRowId" value={rowIdGetter(getRowId, rows)} />
         <Getter name="columns" value={columns} />
         <Getter name="getCellValue" value={cellValueGetter(getCellValue, columns)} />
+        <Getter name="tableHeaderColumnChains" value={tableHeaderColumnChains} />
```
MaximKudriavtsev: Why this getter defined here? Maybe we should define it in the Table plugin like a `tableHeaderRows` getter?

packages/dx-react-grid/src/plugins/table-fixed-columns.jsx
```diff
@@ -62,19 +66,21 @@ export class TableFixedColumns extends React.PureComponent {
...
         <Getter name="tableColumns" computed={tableColumnsComputed} />
+        <Getter name="tableHeaderColumnChains" computed={tableHeaderColumnChainsComputed} />
         <Template
           name="tableCell"
           predicate={({ tableColumn }) => !!tableColumn.fixed}
         >
           {params => (
             <TemplateConnector>
-              {({ tableColumns }) => {
+              {({ tableColumns, tableHeaderColumnChains }) => {
```
MaximKudriavtsev: Currently, this plugin is dependent on the `TableBandHeader` plugin. We should change `pluginDependencies` constant.

```js
const pluginDependencies = [
   { name: 'Table' },
   { name: 'TableBandHeader', optional: true },
 ];
```

packages/dx-react-grid/src/plugins/table-selection.jsx
```diff
@@ -41,6 +45,9 @@ export class TableSelection extends React.PureComponent {
         {showSelectionColumn && (
           <Getter name="tableColumns" computed={tableColumnsComputed} />
         )}
+        {showSelectionColumn && (
```
MaximKudriavtsev: Maybe we should combine this both conditions to one?

MaximKudriavtsev:
If we will keep this new getter, we should update our API reference the `Plugin Developer Reference` section.

packages/dx-react-grid/src/plugins/table-selection.jsx
ushkal: I've tried this initially but it seems that if 2 getters are rendered within React.Fragment then one of these getters stops working

packages/dx-react-grid/src/plugins/table-fixed-columns.jsx
ushkal: I think we should fix it in separate PR, here's a [task](https://trello.com/c/NTirap7m/1285-grid-other-header-plugins-should-be-defined-as-optional-dependency-for-fixed-columns-plugin) in trello

SergeyAlexeev:
Does the new Getter require a particular plugin order? I mean:

```jsx
<TableSelection />
<TableFilterRow />
```

is equal to

```jsx
<TableFilterRow />
<TableSelection />
```

?

And should we add the new Getter to the 'TableRowDetail' plugin?

packages/dx-grid-core/src/plugins/table-band-header/computeds.js
```diff
@@ -21,3 +23,35 @@ export const tableRowsWithBands = (tableHeaderRows, columnBands, tableColumns) =
     .map((row, index) => ({ key: `${TABLE_BAND_TYPE.toString()}_${index}`, type: TABLE_BAND_TYPE, level: index }));
   return [...tableBandHeaders, ...tableHeaderRows];
 };
+
+export const tableHeaderColumnChainsWithBands = (
+  tableHeaderColumnChains, tableHeaderRows, tableColumns, bands,
```
MaximKudriavtsev: The `tableHeaderColumnChains` argument did not used

packages/dx-react-grid/src/plugins/table-band-header.jsx
```diff
@@ -40,6 +45,7 @@ export class TableBandHeader extends React.PureComponent {
         ]}
       >
         <Getter name="tableHeaderRows" computed={tableHeaderRowsComputed} />
+        <Getter name="tableHeaderColumnChains" computed={tableHeaderColumnChainsComputed} />
```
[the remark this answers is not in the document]
ushkal: I think it's unnecessary since the getter became internal: any custom header plugin should not consider it's value or recompute it

PR #1621, opened 2018-11-16
chore(react-chart): do not pass "restProps" down to series component

DmitryBogomolov:
Prevents passing all properties from series plugin component down to series component.


packages/dx-react-chart/src/templates/series/point-collection.jsx
```diff
@@ -20,6 +21,7 @@ export class PointCollection extends React.PureComponent {
     return (coordinates.map(item => (
       <Point
         key={item.index.toString()}
+        seriesIndex={index}
```
Krijovnick: Is `seriesIndex` field needed?

PR #1624, opened 2018-11-19
chore(react-chart): add Line and Spline series components

DmitryBogomolov:
Updates *LineSeries* and *SplineSeries* so that their *Path* components reflect corresponding difference.
Now `SplineSeries.Path` will render curved line.

BREAKING CHANGES:
Starting with this release, the *Area*, *Line*, and *Spline* series do not pass the `path` property to their `seriesComponent` component because it is an inner part of the component's implementation.
You can create *path* function using [d3.line](https://github.com/d3/d3-shape#line) and [d3.area](https://github.com/d3/d3-shape#area).

### DIFF
--- a/packages/dx-react-chart/src/plugins/area-series.jsx
+++ b/packages/dx-react-chart/src/plugins/area-series.jsx
@@ -1,14 +1,11 @@
 import {
-  dArea as path,
-  getAreaPointTransformer as getPointTransformer,
-  createAreaHitTester as createHitTester,
+  getAreaPointTransformer as getPointTransformer, createAreaHitTester as createHitTester,
 } from '@devexpress/dx-chart-core';
 import { declareSeries } from '../utils';
 import { Area as Path } from '../templates/series/area';
 
 export const AreaSeries = declareSeries('AreaSeries', {
   components: { Path },
-  path,
   getPointTransformer,
   createHitTester,
 });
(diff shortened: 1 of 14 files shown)


packages/dx-react-chart/src/plugins/area-series.jsx
```diff
@@ -1,14 +1,11 @@
...
+  getAreaPointTransformer as getPointTransformer, createAreaHitTester as createHitTester,
 } from '@devexpress/dx-chart-core';
 import { declareSeries } from '../utils';
 import { Area as Path } from '../templates/series/area';
 
 export const AreaSeries = declareSeries('AreaSeries', {
   components: { Path },
-  path,
   getPointTransformer,
   createHitTester,
```
Krijovnick: As the `path` property is no longer passes through the `declareSeries` HOC, it is should be remove in that function too.
              `<currentSeries.seriesComponent
                    ...
                    path={currentSeries.path}
                    ...
                  />`

Krijovnick:
The `path` property can be removed from `PointCollection` Component.

Krijovnick:
The `seriesComponent` in area, line and spline series receive not required path that should be specified in the documentation.

packages/dx-react-chart/src/templates/series/area.jsx
```diff
@@ -27,7 +29,7 @@ class RawArea extends React.PureComponent {
 }
 
 RawArea.propTypes = {
-  path: PropTypes.func.isRequired,
+  path: PropTypes.func,
```
Krijovnick: The `seriesComponent` in area, line and spline series receive not required path that should be specified in the documentation.

packages/dx-react-chart/src/plugins/area-series.jsx
DmitryBogomolov: Done.

packages/dx-react-chart/src/templates/series/area.jsx
DmitryBogomolov: Done.

PR #1625, opened 2018-11-19
refactor(react-grid-bootstrap): extract border/background color calculation

MaximKudriavtsev:


packages/dx-react-grid-bootstrap3/src/templates/table-band-header/cell.test.jsx
```diff
@@ -48,16 +35,18 @@ describe('TableCell', () => {
       .toBe('none');
   });
 
-  it('should apply left border if necessary', () => {
-    const tree = shallow((
-      <Cell className="custom-class" beforeBorder />
+  fit('should apply left border if necessary', () => {
```
ushkal: fit -> it

PR #1630, opened 2018-11-20
refactor(react-chart): fix unnecessary re-rendering of PaneLayout plugin

Krijovnick:


packages/dx-react-chart/src/plugins/pane-layout.jsx
```diff
@@ -28,43 +28,36 @@ export class PaneLayout extends React.PureComponent {
   constructor(props) {
     super(props);
 
-    this.state = {
-      width: 800,
-      height: 600,
+    this.handleSizeUpdate = (size, changeBBox) => {
```
DmitryBogomolov: There is no need to keep it as method - it does not use `this`.

PR #1635, opened 2018-11-21
refactor(react-core): replace ref callbacks with createRef

ushkal:


packages/dx-react-chart/src/plugins/axis.jsx
```diff
@@ -158,7 +160,7 @@ class RawAxis extends React.PureComponent {
                     flexGrow: dx || undefined,
                   }}
                   // TODO: *ref* should be created in constructor.
```
DmitryBogomolov: This *TODO* is resolved by this PR.

packages/dx-react-grid-material-ui/src/templates/table-header-cell.jsx
```diff
@@ -147,10 +148,10 @@ class TableHeaderCellBase extends React.PureComponent {
 
     return draggingEnabled ? (
       <DragSource
-        ref={(element) => { this.cellRef = element; }}
+        ref={this.cellRef}
         payload={[{ type: 'column', columnName: column.name }]}
         onStart={() => this.setState({ dragging: true })}
-        onEnd={() => this.cellRef && this.setState({ dragging: false })}
+        onEnd={() => this.cellRef.current && this.setState({ dragging: false })}
```
MaximKudriavtsev: Can we extract this functions from render method? The same in bs3/bs4

MaximKudriavtsev:
Looks like a excess [line](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-grid/src/components/table-layout.jsx#L24)

PR #1649, opened 2018-11-28
chore(react-chart): add boolean properties to enable/disable components in Axis

Krijovnick:


packages/dx-react-chart/src/plugins/axis.jsx
```diff
@@ -159,7 +164,7 @@ class RawAxis extends React.PureComponent {
           <TemplateConnector>
             {({ scales, layouts }) => {
               const scale = scales[name];
-              if (!scale) {
+              if (!scale || showGrids === false) {
```
DmitryBogomolov: Why not `!showGrids`?

Krijovnick: done

PR #1656, opened 2018-11-29
refactor: get rid of reactstrap dependency

### LINKED ISSUE #1543: Request: Reduce bundle size by removing "reactstrap" dependency

### I'm using ...

- [x] React Grid
- [ ] React Chart

### Current Behaviour

The grid with Bootstrap 4 uses "reactstrap" which increases the bundle size by 85KB.

### Expected Behaviour

Remove the dependency on "reactstrap" because almost none of it is used.

### More information

I am using Bootstrap 4 in my React application without "reactstrap".

I checked the source of the DevExtreme React Grid and only four components of "reactstrap" are used:

`import { Popover, Pagination, PaginationItem, PaginationLink } from 'reactstrap';`

Everything else is implemented by plain HTML with Bootstrap classes.

I believe it does not make sense to include "reactstrap" for these four components. The pagination components are just "sugar" around the Bootstrap 4 pagination HTML and classes. This could easily be replaced.

The Popover in "reactstrap" is Bootstrap 4 HTML with "react-popper". So, why not use "react-popper" directly and make "react-popper" a peer dependency instead?

With these minor changes the bundle size would decrease and "reactstrap" would be optional. And if someone would want to use "reactstrap", there would not be an issue because both libraries would depend on the same "react-popper" and the bundle size would not increase.

Here is an overview of the bundle size of a simple grid sample:

gridsamplebundlesize

You can see that "reactstrap" accounts for 20% of the bundle size.

### Environment

- devextreme-reactive: 1.8.0
- react: 16.6
- browser: All
- bootstrap: 4
- react-bootstrap: none
- material-ui: none

ushkal:
closes #1543


packages/dx-react-grid-bootstrap4/src/templates/column-chooser/overlay.test.jsx
```diff
@@ -35,4 +36,40 @@ describe('Overlay', () => {
     expect(tree.props().data)
       .toMatchObject({ a: 1 });
   });
+
+  describe('onHide', () => {
+    it('should be called on toggle if overlay is vidible', () => {
```
SergeyAlexeev: vidible -> visible

packages/dx-react-grid-bootstrap4/src/templates/column-chooser/overlay.test.jsx
```diff
@@ -35,4 +36,40 @@ describe('Overlay', () => {
...
+          <div />
+        </Overlay>
+      ));
+
+      tree.find(Popover).prop('toggle')();
+
+      expect(onHide).toHaveBeenCalled();
+    });
+
+    it('should not be called on toggle if overlay is not vidible', () => {
```
SergeyAlexeev: The same about 'vidible'.

packages/dx-react-bootstrap4/jest.config.js
```diff
@@ -0,0 +1,11 @@
+const path = require('path');
+
+module.exports = {
+  verbose: true,
+  transform: {
+    '^.+\\.(js|jsx)?$': 'babel-jest',
+  },
```
MaximKudriavtsev: Is `transform` required? [Note](https://jestjs.io/docs/en/getting-started#using-webpack)

> `babel-jest` is automatically installed when installing Jest and will automatically transform files if a babel configuration exists in your project.

packages/dx-react-chart-bootstrap4/src/templates/tooltip/overlay.jsx
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import { Popover, PopoverBody } from 'reactstrap';
+import { Popover } from '../../../../dx-react-bootstrap4/components/popover';
```
MaximKudriavtsev: You can change this path to `../../../../dx-react-bootstrap4/components` because the `index.js` file exists in a `components` filder.

PR #1666, opened 2018-11-30
chore(react-chart): fix title position

Krijovnick:


packages/dx-react-chart-demos/src/demo-sources/bar-chart/material-ui/series-interaction.jsx
```diff
@@ -68,11 +68,17 @@ const contentBase = ({ classes, ...props }) => {
 };
 const ContentComponent = withStyles(contentStyles, { name: 'contentStyles' })(contentBase);
 
+const demoStyles = () => ({
+  title: {
+    marginRight: '120px',
```
DmitryBogomolov: Let's pass it as inline style (like in *bootstrap4* counterpart).

PR #1670, opened 2018-12-03
chore(react-chart): add event data to the event handlers

Krijovnick:


packages/dx-react-chart-demos/src/demo-sources/interaction/single-multiple-selection.jsxt
```diff
@@ -0,0 +1,83 @@
...
+      if (target) {
+        this.setState(({ selection }) => {
+          if (event.ctrlKey) {
+            const filtered = selection.filter(el => !compare(el, target));
+            return {
+              selection: filtered.length < selection.length ? filtered : [...filtered, target],
+            };
+          }
+          return {
+            selection: selection[0] && compare(selection[0], target) ? [] : [target],
```
DmitryBogomolov: Let's say there are 3 selected items by this moment. If one of them is clicked, selection is reset. If any of others - only that item is selected. Inconsistency.

When item is clicked all other items selection is reset and this item's selection is toggled.

PR #1673, opened 2018-12-04
chore(react-chart): update multiple axes demo

DmitryBogomolov:
Adds more series, hover and tooltip to *combination/multiple-axes* demo.


DmitryBogomolov:
Here is how it looks.

combination-axes_bootstrap4
combination-axes_materialui

Krijovnick:
Maybe all series should be hovered? It is very strange that only one series hovered and show tooltip.

DmitryBogomolov:
Tooltip in *for-all-series* mode has been restored.

PR #1674, opened 2018-12-04
fix(react-grid): prevent animation flicker on column visibility change

ushkal:


packages/dx-react-grid/src/components/table-layout.jsx
```diff
@@ -19,19 +19,37 @@ export class TableLayout extends React.PureComponent {
...
+    const tableWidth = this.getTableWidth(prevColumns, columns);
     this.animations = getAnimations(prevColumns, columns, tableWidth, this.animations);
+
     cancelAnimationFrame(this.raf);
     this.raf = requestAnimationFrame(this.processAnimationFrame.bind(this));
   }
 
+  getTableWidth(prevColumns, columns) {
+    // eslint-disable-next-line react/no-find-dom-node
+    const { offsetWidth, scrollWidth } = findDOMNode(this);
```
MaximKudriavtsev: Can we avoid to use a `findDOMNode` function? Because `strictMode` said that this is a deprecated API.

ushkal: Yes, we can try to do it since tableLayout is not public

packages/dx-react-grid/src/components/table-layout.jsx
```diff
@@ -19,19 +18,39 @@ export class TableLayout extends React.PureComponent {
...
-    const tableWidth = findDOMNode(this).scrollWidth;
+    const tableWidth = this.getTableWidth(prevColumns, columns);
     this.animations = getAnimations(prevColumns, columns, tableWidth, this.animations);
+
     cancelAnimationFrame(this.raf);
     this.raf = requestAnimationFrame(this.processAnimationFrame.bind(this));
   }
 
+  getTableWidth(prevColumns, columns) {
+    // eslint-disable-next-line react/no-find-dom-node
```
SergeyAlexeev: Looks like it's an unnecessary comment.

packages/dx-react-grid-material-ui/src/templates/table.jsx
```diff
@@ -3,6 +3,9 @@ import * as PropTypes from 'prop-types';
 import classNames from 'classnames';
 import TableMUI from '@material-ui/core/Table';
 import { withStyles } from '@material-ui/core/styles';
+// eslint-disable-next-line no-restricted-imports
+import { RootRef } from '@material-ui/core';
```
SergeyAlexeev: Why don't we use the [recommended](https://material-ui.com/guides/minimizing-bundle-size/#how-to-reduce-the-bundle-size) import statement `import RootRef from '@material-ui/core/RootRef';`?

PR #1677, opened 2018-12-05
chore(react-core): fix the Action plugin unmount behavior

MaximKudriavtsev:


packages/dx-react-core/src/plugin-based/getter.test.jsx
```diff
@@ -206,4 +206,17 @@ describe('Getter', () => {
     expect(log).toHaveLength(2);
     expect(log[0]).not.toBe(log[1]);
   });
+
+  it('should correct Getter unmount', () => {
+    const value = 0;
+
+    const wrapper = mount((
+      <PluginHost>
+        <Getter name="action" value={value} />
```
SergeyAlexeev: name="action" => name="getter" 😉

PR #1678, opened 2018-12-05
chore(react-chart): update continuous series hit testing

DmitryBogomolov:
Selects point closest to pointer position when *area*, *line*, *spline* series is hovered. Now hover target always has `point` field.

### DIFF
--- a/packages/dx-react-chart/src/plugins/selection-state.jsx
+++ b/packages/dx-react-chart/src/plugins/selection-state.jsx
@@ -9,7 +9,8 @@ import { changeSeriesState, SELECTED } from '@devexpress/dx-chart-core';
 export class SelectionState extends React.PureComponent {
   render() {
     const { selection } = this.props;
-    const getSeries = ({ series }) => changeSeriesState(series, selection, SELECTED);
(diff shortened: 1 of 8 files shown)


packages/dx-react-chart/src/plugins/selection-state.jsx
```diff
@@ -21,10 +22,10 @@ export class SelectionState extends React.PureComponent {
 SelectionState.propTypes = {
   selection: PropTypes.arrayOf(PropTypes.shape({
     series: PropTypes.string.isRequired,
-    point: PropTypes.number,
+    point: PropTypes.number.isRequired,
```
Krijovnick: What about if I want select/hover only series, without point?

Krijovnick: The documentation should be fix, too

DmitryBogomolov: It is in another [branch](https://github.com/DmitryBogomolov/devextreme-reactive/tree/update-hover-docs).

PR #1685, opened 2018-12-07
fix(react-chart): fix resizing of the axes

Krijovnick:


packages/dx-react-chart/src/templates/axis/root.jsx
```diff
@@ -26,18 +25,15 @@ export class Root extends React.PureComponent {
...
     const bbox = this.node.getBBox();
     const width = dx ? bbox.width : getSize(bbox.x, bbox.width);
     const height = dy ? bbox.height : getSize(bbox.y, bbox.height);
     const x = dx ? 0 : getOffset(bbox.x);
     const y = dy ? 0 : getOffset(bbox.y);
-    if (width === prevWidth && height === prevHeight) {
-      return null;
-    }
     onSizeChange({ width, height });
     return {
```
DmitryBogomolov: Since now new state does not depend on current state, callback version of `setState` is not required.
This line can then be replaced with `this.setState({ x, y })` and in `componentDidMount` and `componentDidUpdate` those single lines can be resplaced with `this.adjust()`;
There is comment over `componentDidUpdate` - it should be update. Now unconditional `setState` relies on the "pureness" of the component.

PR #1693, opened 2018-12-10
fix(react-grid): specify table fixed columns plugin optional dependencies

ushkal:


MaximKudriavtsev:
Should we add this list of plugins to our API reference?

packages/dx-react-grid/src/plugins/table-fixed-columns.jsx
```diff
@@ -20,6 +20,14 @@ const CellPlaceholder = props => <TemplatePlaceholder params={props} />;
...
 const pluginDependencies = [
   { name: 'Table' },
+  { name: 'TableBandHeader', optional: true },
+  { name: 'TableColumnReordering', optional: true },
+  { name: 'TableEditColumn', optional: true },
+  { name: 'TableEditRow', optional: true },
+  { name: 'TableFilterRow', optional: true },
+  { name: 'TableHeaderRow', optional: true },
+  { name: 'TableSelection', optional: true },
+  { name: 'TableTreeColumn', optional: true },
```
MaximKudriavtsev: What about the `TableSummaryRow` plugin? Next behavior looks like a bug

image

PR #1695, opened 2018-12-11
chore(react-chart): improve small targets tracking

DmitryBogomolov:
Removes complex hover target picking logic. Now point belonging to the last declared series has priority unless its distance is not much greater than that of other candidates.

Tracker now operates with absolute distances rather than relative (like it did beforehand).

### DIFF
--- a/packages/dx-chart-core/src/utils/event-tracker.js
+++ b/packages/dx-chart-core/src/utils/event-tracker.js
@@ -11,7 +11,7 @@ const getEventCoords = (e) => {
 
 const DISTANCE_THRESHOLD = 20;
 
-const compare = (t1, t2) => {
+const compareHitTargets = (t1, t2) => {
   const distanceDelta = t1.distance - t2.distance;
   if (Math.abs(distanceDelta) <= DISTANCE_THRESHOLD) {
     const orderDelta = t2.order - t1.order;
@@ -45,7 +45,7 @@ const buildEventHandler = (seriesList, handlers) => {
         ));
       }
     });
-    targets.sort(compare);
+    targets.sort(compareHitTargets);
     const arg = { location, targets, event: e.nativeEvent };
     handlers.forEach(handler => handler(arg));
   };
--- a/packages/dx-chart-core/src/utils/event-tracker.js
+++ b/packages/dx-chart-core/src/utils/event-tracker.js
@@ -9,7 +9,16 @@ const getEventCoords = (e) => {
   ];
 };
 
-const compare = (t1, t2) => t1.distance - t2.distance;
+const DISTANCE_THRESHOLD = 20;
+
+const compare = (t1, t2) => {
+  const distanceDelta = t1.distance - t2.distance;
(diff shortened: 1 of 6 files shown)


packages/dx-chart-core/src/utils/event-tracker.js
```diff
@@ -9,7 +9,16 @@ const getEventCoords = (e) => {
   ];
 };
 
-const compare = (t1, t2) => t1.distance - t2.distance;
+const DISTANCE_THRESHOLD = 20;
+
+const compare = (t1, t2) => {
```
MaximKudriavtsev: I think `compare` is a very common name. Very difficult to search where it used.

packages/dx-chart-core/src/utils/series.js
```diff
@@ -1,15 +1,7 @@
...
-// Based on https://en.wikipedia.org/wiki/Uniform_norm
-const getUniformDistance = (dx, dy, rx, ry) => Math.max(
-  Math.abs(dx) / rx, Math.abs(dy) / ry,
-);
-
-// Based on https://en.wikipedia.org/wiki/Norm_(mathematics)#Euclidean_norm
-const getEuclideanDistance = (dx, dy, rx, ry) => Math.sqrt(
-  (dx * dx) / (rx * rx) + (dy * dy) / (ry * ry),
-);
+const getLength = (dx, dy) => Math.sqrt(dx * dx + dy * dy);
```
MaximKudriavtsev: Maybe we should rename this function to `getDistance` for example? Because of the `length` word used for arrays and string and return a number of values, but in current place, we would like to see a distance between two points.

`getRadius`, `getHypotenuse`

packages/dx-chart-core/src/utils/event-tracker.js
DmitryBogomolov: More specific name is applied.

packages/dx-chart-core/src/utils/series.js
DmitryBogomolov: Renamed.

packages/dx-chart-core/src/utils/series.js
```diff
@@ -134,9 +129,13 @@ export const createPieHitTester = createPointsEnumeratingHitTesterCreator(
     const halfAngle = Math.abs(startAngle - endAngle) / 2;
     const dx = px - x;
     const dy = py - y;
-    const r = Math.sqrt(dx * dx + dy * dy);
+    const r = getSegmentLength(dx, dy);
     const angle = mapAngleTod3(Math.atan2(dy, dx));
-    return getUniformDistance(r - rCenter, angle - angleCenter, halfRadius, halfAngle);
+    return Math.abs(r - rCenter) <= halfRadius && Math.abs(angle - angleCenter) <= halfAngle ? {
```
ushkal: This return code logic looks the same as in createBarHitTester()

DmitryBogomolov: Resolved.

PR #1719, opened 2018-12-19
fix(react-scheduler): add to AllDayPanel another cell template name

MaximKudriavtsev:


SergeyAlexeev:
What about a test?

I suggest replacing this line:

```
layoutComponent: ({ children }) => <div className="layout">{children}</div>
```

to 

```
const LayoutComponent = () => null;
layoutComponent: LayoutComponent,
```

In this case, you can access all layout component props and check them.

BTW:

It looks like there are two similar tests. 'should render AllDayPanel layout' and 'should render day panel'.

They both check that the `.layout` component exists.

PR #1724, opened 2018-12-21
chore(react-chart): update multiple-axes demo

DmitryBogomolov:


MaximKudriavtsev:
The title align is not center.

image

PR #1726, opened 2018-12-21
chore(site): beautify source code

MaximKudriavtsev:


packages/dx-demo-shell/src/utils.js
```diff
@@ -0,0 +1,6 @@
+export const findOccurrence = (source, strings) => {
+  for (let index = 0; index < strings.length; index += 1) {
+    if (source.indexOf(strings[index]) > -1) return true;
+  }
+  return false;
+};
```
SergeyAlexeev: [Array.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) ?

packages/dx-demo-shell/src/utils.js
```diff
@@ -0,0 +1,6 @@
+export const findOccurrence = (source, strings) => {
```
DmitryBogomolov: `(source, strings) => strings.some(str => source.includes(str))`

And do we need additional file because of single helper function?

packages/dx-demo-shell/src/demo-viewer/source-code.jsx
```diff
@@ -34,16 +38,19 @@ export class SourceCode extends React.PureComponent {
...
-    });
-    this.importantLines.forEach((lineNumber) => {
-      editor.addLineClass(lineNumber, 'background', 'CodeMirror-important-line');
-    });
+    this.applySpecialCodeOptions();
   }
 
-  prepareSourceCode() {
-    const { themeName, sectionName, demoName } = this.props;
+  componentWillReceiveProps(nextProps) {
```
DmitryBogomolov: It is [obsolete](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops).

packages/dx-demo-shell/src/demo-viewer/source-code.jsx
```diff
@@ -64,12 +72,22 @@ export class SourceCode extends React.PureComponent {
...
+    });
+    this.importantLines.forEach((lineNumber) => {
+      this.codeMirror.addLineClass(lineNumber, 'background', 'CodeMirror-important-line');
+    });
+  }
+
   render() {
-    const sourceCode = this.prepareSourceCode();
+    const { themeName, sectionName, demoName } = this.props;
+    const sourceCode = this.prepareSourceCode(themeName, sectionName, demoName);
```
DmitryBogomolov: Even though now `foldBlockStartLines` and `importantLines` are reset on each `render` it is still not pure as modifying instance fields is a side effect.

However it will most like be automatically resolved when `componentWillReceiveProps` is gone.

packages/dx-demo-shell/src/utils.js
MaximKudriavtsev: Do you suggest use `array.find()` instead of `array.findIndex()`?

ushkal: Maybe `strings.some(s => source.indexOf(s) > -1)` ?

packages/dx-demo-shell/src/demo-viewer/source-code.jsx
```diff
@@ -34,16 +38,19 @@ export class SourceCode extends React.PureComponent {
...
-      editor.foldCode(CodeMirror.Pos(lineNumber, 0));
-    });
-    this.importantLines.forEach((lineNumber) => {
-      editor.addLineClass(lineNumber, 'background', 'CodeMirror-important-line');
-    });
+    this.applySpecialCodeOptions();
+  }
+
+  componentDidUpdate() {
+    const sourceCode = this.prepareSourceCode();
```
DmitryBogomolov: `prepareSourceCode` is called twice after component is mount - in `render` and here.

packages/dx-demo-shell/src/demo-viewer/source-code.jsx
```diff
@@ -64,12 +72,21 @@ export class SourceCode extends React.PureComponent {
...
+    this.foldBlockStartLines.forEach((lineNumber) => {
+      this.codeMirror.foldCode(CodeMirror.Pos(lineNumber, 0));
+    });
+    this.importantLines.forEach((lineNumber) => {
+      this.codeMirror.addLineClass(lineNumber, 'background', 'CodeMirror-important-line');
+    });
+  }
+
   render() {
     const sourceCode = this.prepareSourceCode();
```
DmitryBogomolov: Can't the following be applied?

```js
static getDerivedStateFromProps(props) {
  return {
    sourceCode: <calculate-from-props>
    foldBlockStartLines: <calculate-from-props>
    importantLines: <calculate-from-props>
  }
}

updateCodeMirror() {
  this.codeMirror.setValue(this.state.sourceCode);
  <apply-special-code-options>
}

componentDidMount() {
  <create-code-mirror-instance>
  this.updateCodeMirror();
}

componentDidUpdate() {
  this.updateCodeMirror();
}

render() {
  return <textarea ref={this.textarea} onChange={handleChange} />;
}
```

packages/dx-demo-shell/src/demo-viewer/source-code.jsx
```diff
@@ -34,42 +37,61 @@ export class SourceCode extends React.PureComponent {
...
   }
 
   render() {
-    const sourceCode = this.prepareSourceCode();
     return (
       <textarea
         ref={this.textarea}
-        value={sourceCode}
+        defaultValue=""
         onChange={() => {}}
```
DmitryBogomolov: Extract `() => {}` to a constant. Now it falsely forces `textarea` to update.

packages/dx-demo-shell/src/demo-viewer/source-code.jsx
```diff
@@ -34,42 +37,61 @@ export class SourceCode extends React.PureComponent {
...
-    });
-    this.importantLines.forEach((lineNumber) => {
-      editor.addLineClass(lineNumber, 'background', 'CodeMirror-important-line');
-    });
+    this.codeMirror.setValue(source);
+    this.applySpecialCodeOptions(foldBlockStartLines, importantLines);
+  }
+
+  componentDidUpdate() {
+    const { source, foldBlockStartLines, importantLines } = this.prepareSourceCode();
```
DmitryBogomolov: Extract these three lines to a separated method. Call it in `componentDidMount` then.

packages/dx-demo-shell/src/utils.js
rao123dk: `Array.includes() `?

PR #1731, opened 2018-12-25
fix(react-grid): prevent tree column header text from overflowing a container

ushkal:


packages/dx-react-grid-bootstrap4/src/templates/table-header-cell/content.jsx
```diff
@@ -3,14 +3,18 @@ import * as PropTypes from 'prop-types';
...
   <div
     className={classNames({
       'w-100 d-flex flex-row align-items-end': true,
       'justify-content-center': align === 'center',
       'justify-content-end': align === 'right',
     }, className)}
+    style={{
+      minWidth: 0,
+      ...style,
+    }}
```
MaximKudriavtsev: We should avoid inline styles in bs4 / mui packages. We can create custom classes with these styles in the `dx-styles` package and apply class instead of styles.

packages/dx-react-grid-material-ui/src/templates/table-header-cell/content.jsx
```diff
@@ -20,14 +20,18 @@ const styles = {
...
 }) => (
   <div
     className={classNames({
       [classes.content]: true,
       [classes.alignCenter]: align === 'center',
       [classes.alignRight]: align === 'right',
     }, className)}
+    style={{
+      minWidth: 0,
+      ...style,
```
MaximKudriavtsev: Use className instead

PR #1737, opened 2018-12-27
docs(react-grid): add plugin development guide

NickMitrokhin:


MaximKudriavtsev:
Can we make a table of contents for this guide? And change these part names to more common titles.

For example
- Rendering Custom Markup - ok
- Providing Values For The Select Box -> Providing Values Between Plugins
- Filtering The Grid Based On A Selected Value -> Combine functionality built-in plugins with custom plugins (?) / Using built-in plugins functionality
- Clearing The Current Filter Value And Displaying The Current Column Caption -> something like - Final Clearing / Finals /

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+export default {
+  getCustomers() {
+    return customers;
+  }
+}
+```
+
+Let's take a look at the screenshot below to see the final result that we will get after performing all the steps from the tutorial:
+
+![](../../../../../img/guides/custom-plugin-development/result.png)
```
MaximKudriavtsev: I suggest using a `html` for pass images because we should add specials styles for good show these for other devices.

I suggest adding a class 'img-responsive' for resolve problems described above.

```jsx
<p align="center">
  <img class="img-responsive" src="img-path.img.png">
</p>
```

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+    return (
+      <Plugin
+        name="ToolbarFilter"
+      >
+      </Plugin>
+    );
+  }
+}
+```
+We are going to show our plugin in the Grid's [toolbar](../reference/toolbar.md). According to the documentation, the toolbar exports the `toolbarContent` template. So, we need to use this template in our plugin for modifying its content. For this task we need to add the [Template](../../../core/docs/reference/template.md) component to our plugin and set its `name` property to `toolbarContent`.
```
MaximKudriavtsev: ... toolbar exports a [Template]() with `toolbarContent` name.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+        <Template name="toolbarContent">
+          <TemplatePlaceholder />
+        </Template>
+      </Plugin>
+    );
+  }
+}
+```
+
+The plugin should not work without the toolbar, let's notify the plugin about it by specifying the `dependencies` option as shown below:
```
MaximKudriavtsev: ... without the `Toolbar` plugin because we have started use its template.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+  }
+}
+```
+After that, our grid should look as follows:
+![](../../../../../img/guides/custom-plugin-development/add-select.png)
+
+## Providing Values For The Select Box
+We see our select box, but it does nothing. The next step we should perform is to add items to our select box. These items should represent unique values of a particular column by which we want to filter our grid. The `ToolbarFilter` plugin represents a UI [plugin](./plugin-overview.md). This plugin should only render content based on provided data. But it should not manipulate the Grid's state or modify data. For this task, we need to implement an extra plugin called [state management](./plugin-overview.md) plugin. This plugin will prepare all necessary information for our `ToolbarFilter` plugin. Let's name it `ToolbarFilterState`. So, add a new `file toolbar-filter-state.js` file to the `plugins` folder of our project and implement the logic of a new plugin in it.
+
+*file toolbar-filter-state.js:*
```
MaximKudriavtsev: Why you used the `file` prefix with only `toolbar-filter-state.js`?  Maybe remove the `file` word?

MaximKudriavtsev:
Can we add to the `codesandbox` example more rows with same company names, this show that filter can display not only one row.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+} from '@devexpress/dx-react-core';
+
+export class ToolbarFilter extends React.PureComponent {
+  render() {
+    return (
+      <Plugin
+        ...
+      >
+        <Template name="toolbarContent">
+          </Template>
```
SergeyAlexeev: Unnecessary spaces.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+} from '@devexpress/dx-react-core';
+
+export class ToolbarFilterState extends React.PureComponent {
+  constructor(props) {
+    super(props);
+        ...
+        this.state = {
+            filterValue: this.props.defaultFilterValue
+        };
+        this.changeValue = this.changeValue.bind(this);
```
SergeyAlexeev: Extra spaces.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+
+![](../../../../../img/guides/custom-plugin-development/result.png)
+
+As you can see, there is a select box at the top right corner of the grid associated with the `CompanyName` column. When a user selects an item from the box, the grid displays only filtered items. Moreover, there is the `CLEAR` button that clears the selected value. When the value is cleared, the grid displays all available rows.
+
+
+Let's start extending our grid!
+
+## Rendering Custom Markup
+For rendering custom markup in existing grid UI elements, in our case it is the toolbar, we need to create a custom plugin. For creating a plugin, we need to wrap all content in the [Plugin](../../../core/docs/reference/plugin.md) component. Since our plugin will reside in the grid's toolbar, we will name it `ToolbarFilter`. Let's create a new folder `plugins` in our project and
```
ushkal: I think 'To create a plugin' sounds better considering the previous sentence with same construction

packages/dx-react-grid/docs/guides/custom-plugin-development.md
ushkal: Can be just <Template name="toolbarContent" **/**>

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+    };
+    ...
+  }
+
+  render() {
+    ...
+  }
+}
+```
+In our scenario, we are going to filter the grid by the `CompanyName` column. The default filter value should be set to an empty string which means that the grid should not fitler items initially. So, we can add this plugin to our grid. Since it is a state management plugin it should included before UI plugins. See [React Grid - Plugin Overview](./plugin-overview.md).
```
ushkal: should **be** inlcuded

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+      >
+        <Template name="toolbarContent">
+          <TemplatePlaceholder />
+          <TemplateConnector>
+          {({
+              toolbarFilterDataItems
+            }) => (
+              <Select>
+                {toolbarFilterDataItems.map((item, index) =>
+                  <MenuItem key={index} value={item}>{item}</MenuItem>
```
ushkal: It's better to use `item` instead of `index` as a key

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+class App extends Component {
+  ...
+  render() {
+    ...
+    return (
+      <Paper>
+        <Grid
+          ...
+        >
+          <ToolbarFilterState defaultColumnName={"CompanyName"} defaultFilterValue={""}/>
```
ushkal: These curly braces can be removed since we pass string literals: 
`defaultColumnName="CompanyName"`

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+    );
+  }
+}
+```
+
+*file toolbar-filter-state.js:*
+```jsx
+...
+
+const filterExpressionComputed = ({filterExpression, toolbarFilterValue, toolbarFilterColumnName}) => {
```
ushkal: Need spaces inside curly braces

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,687 @@
...
+      </Plugin>
+    );
+  }
+}
+```
+
+*toolbar-filter.js:*
+```jsx
+...
+const columnTitle = function(columns, toolbarFilterColumnName) {
```
ushkal: I suggest to rename `columnTitle` -> `getColumnTitle`

SergeyAlexeev: Also, we prefer '() => {}' instead of 'function';

packages/dx-react-grid/docs/guides/custom-plugin-development.md
NickMitrokhin: It was included by mistake. I'll remove it.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
NickMitrokhin: I'm not sure I understand in what part of the text you wish to use 'To create plugin'. Would you please clarify?

ushkal: Instead of 'For creating a plugin'

packages/dx-react-grid/docs/guides/custom-plugin-development.md
NickMitrokhin: I think it doesn't matter in this particular case.

ushkal: It's just a React best practice: using a loop index as a key is discouraged

NickMitrokhin:
@MaximKudriavtsev a table of contents is a good idea. I'll add it and rename sections. As for adding extra rows, this is just an example, I think we should not care about a number of data rows.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
NickMitrokhin: Ok, thanks

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,708 @@
...
+
+*index.js:*
+```jsx
+import {
+  Grid,
+  Table,
+  TableHeaderRow,
+  Toolbar
+} from '@devexpress/dx-react-grid-material-ui';
+import service from './data.js';
```
dxbykov: The `rows: service.getCustomers()` code below looks a bit strange to me since I expect an async call here. So I would replace this with a simpler code:

```js
import customers from './data'
...
{
  ...
   rows: customers
}
```

NickMitrokhin: Makes sense. I'll fix it.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,708 @@
...
+}
+```
+After you've done that, the Grid should look as follows:
+<p align="center">
+  <img class="img-responsive" src="../../../../../img/guides/custom-plugin-development/add-select.png">
+</p>
+
+## Share Values Between Plugins
+
+We see our select box, but it does nothing. The next step we should perform is to add items to our select box. These items should represent unique values of a particular column by which we want to filter our grid. The `ToolbarFilter` plugin represents a UI [plugin](./plugin-overview.md). This plugin should only render content based on provided data. But it should not manipulate the Grid's state or modify data. For this task, we need to implement an extra plugin called [state management](./plugin-overview.md) plugin. This plugin will prepare all necessary information for our `ToolbarFilter` plugin. Let's name it `ToolbarFilterState`. So, add a new `toolbar-filter-state.js` file to the `plugins` folder of our project and implement the logic of a new plugin in it.
```
dxbykov: I would mention here that the UI/State plugin separation is optional. We can implement all stuff in a single plugin, but the separation of concerns make it more flexible and maintainable.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,708 @@
...
+      <Plugin
+        name="ToolbarFilterState"
+      >
+      </Plugin>
+    );
+  }
+}
+```
+
+Now we need to specify by which column we wish to filter our grid. The "columnName" option is appropriate for this task. Since we are not going to implement [controlled](./controlled-and-uncontrolled-modes.md) mode for this plugin add the `default` prefix to the property name according to the naming convention. So, the final name will be `defaultColumnName`.
```
dxbykov: The `default` prefix means that a user specifies the *initial* value for the component state property allowing the component to change it further internally (the user doesn't control the component state).  In our case, the columnName property is not a part of the component state. It's not changed internally. So, the `default` prefix doesn't make sense here.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,702 @@
...
+
+In addition, to the `defaultColumnName` property, let's add the `defaultFilterValue` property to allow a user to apecify initial filter value. When a user selects a value in our select box, this value will be changed. In this case, save it in the state object of our component as shown below:
```
dxbykov: The `default` prefix explanation should go here.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,702 @@
...
+
+How can we get values of a particular column? We set data items using the `rows` property of the [Grid](../reference/grid.md) plugin. This plugin exports the `rows` getter. We can use it in our case. However, we will need to pass data items to our UI plugin (`ToolbarFilter`). So, we will declare a new getter for this task and name it `toolbarFilterDataItems`. The value of this getter will be calculated based on the value of the `rows` getter. For this task we will need to use computed functions. These functions calculate values based on values of other getters. As you remember, we will need not only filter our grid, but display the filtered column name near the select box. Since we are going to do that in the `ToolbarFilter` plugin we will need to get the column name in the plugin. A new getter will help us do this. Let's name it `toolbarFilterColumnName`. Moreover, we will use this getter for calculating a value for `toolbarFilterDataItems`. Add a new calculated function named, for example, `filterDataItemsComputed` that accepts the `rows` and `toolbarFilterColumnName` getters and calculates items for our select box. Our code might be as follows:
```
dxbykov: > For this task we will need to use computed functions.

It sounds like the functions are computed. Actually, the value of a getter is computed using the values of other getters. Can we paraphrase this somehow? For instance,  "For this task, we will use functions that we internally call 'computeds'".

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,702 @@
...
+The event handler passes the `event` parameter. We can get a new value using this parameter and pass it to the state object of the `ToolbarFilterState` plugin.
+
+*file toolbar-filter-state.js:*
+```jsx
+...
+export class ToolbarFilterState extends React.PureComponent {
+  ...
+  changeValue(event) {
+    this.setState({
+      filterValue: event.target.value
```
dxbykov: A state plugin shouldn't know anything about the UI specifics. So, the `changeValue` action should accept a ready-to-use filter value, not a UI event.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,702 @@
...
+```
+
+*file toolbar-filter-state.js:*
+```jsx
+...
+
+const filterExpressionComputed = ({ filterExpression, toolbarFilterValue, toolbarFilterColumnName }) => {
+    const newFilterExpression = { ...filterExpression || { filters: [], operator: "and" } };
+    if (toolbarFilterValue) {
+      newFilterExpression.filters.push({
```
dxbykov: We must keep values immutable.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,702 @@
...
+  constructor(props) {
+    super(props);
+    ...
+    this.columnName = this.props.defaultColumnName || null;
+    this.clearValue = this.clearValue.bind(this);
+  }
+
+  clearValue() {
+    this.setState({
+      filterValue: ""
```
dxbykov: We usually use single quotes for string values.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,702 @@
...
+          {({
+              ...,
+              toolbarFilterColumnName,
+              columns
+            }, {
+              ...,
+              clearToolbarFilterValue
+            }) => (
+              <div>
+                <InputLabel htmlFor="filter-field">{getColumnTitle(columns, toolbarFilterColumnName)}:</InputLabel>
```
dxbykov: There is no need to recalculate a column title each time filter is being rendered if the column name hasn't changed. It's enough to calculate it just one time when the column name has changed. So, it should be a computed exposed from the ToolbarFilterState plugin. I would combine it with columnName and export the filter column info (name + title).

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+1. [Introduction](#introduction)
+2. [Render Custom Markup](#render-custom-markup)
+3. [Share Values Between Plugins](#share-values-between-plugins)
+4. [Use the Functionality of Built-In Plugins](#use-the-functionality-of-built-in-plugins)
+5. [Final Steps](#final-steps)
+
+
+## Introduction
+
+This tutorial is aimed at advanced users who are familiar with [basic Grid concepts](./getting-started.md) and who want to add new functionality to the Grid by using the [React Core](../../../core/docs/guides/fundamentals.md) components.
```
dirk-pieterse: Is it necessary to add "This tutorial is aimed at advanced users who are familiar with [basic Grid concepts](./getting-started.md)"?

dirk-pieterse: Why 'by using'?

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+
+In this tutorial, we add to the Grid's toolbar a select box that filters the grid by a predefined column. The purpose of the tutorial is not to implement a valuable feature but to show how to extend the Grid's functionality.
```
dirk-pieterse: This tutorial... In this tutorial...?

dirk-pieterse: "is not to implement a valuable feature"? a valuable feature? Why do you add this?

dirk-pieterse: You've already explained before that this tutorial is about adding functionality. Why do you repreat this?

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+
+We start with the following code. As you can see, we use the Grid in the `Material-UI` theme.
```
dirk-pieterse: Again, a very strange way to explain a simple idea. "We start with the following code. As you can see, we use the Grid in the `Material-UI` theme." -> In the following code, we use the Grid in the `Material-UI` theme:

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+    ID: 5,
+    CompanyName: 'E-Mart',
+    Address: '3333 Beverly Rd',
+    City: 'Hoffman Estates',
+    State: 'Illinois',
+    Zipcode: 60179
+  }];
+```
+
+Take a look at the image below to see what you will get by the end of the tutorial. At the top right corner, you can see a select box. When a user selects a value from it, the grid displays only the filtered rows. When the value is cleared using the `CLEAR` button, the grid displays all rows.
```
dirk-pieterse: The image below shows the final result. There is a select box at the top right corner. When a user selects a value from it, the grid displays only the filtered rows. When the value is cleared using the `CLEAR` button, the grid displays all rows.

dirk-pieterse: Why do you need to use future tense?

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+
+<p align="center">
+  <img class="img-responsive" src="../../../../../img/guides/custom-plugin-development/result.png">
+</p>
+
+The first step is to render custom markup.
+
+## Render Custom Markup
+
+To render custom markup in a UI element that belongs to the Grid ([toolbar](../reference/toolbar.md) in our case), we need to create a plugin. Wrap the custom markup in the [Plugin](../../../core/docs/reference/plugin.md) component to do this. We will name the plugin `ToolbarFilter`, since it will be part of the toolbar and will contain a filtering UI element.
```
dirk-pieterse: will... will... will...? Why?

dirk-pieterse: "To render custom markup in a UI element that belongs to the Grid ([toolbar](../reference/toolbar.md) in our case), we need to create a plugin. Wrap the custom markup in the [Plugin](../../../core/docs/reference/plugin.md) component to do this." -> To render custom markup in a UI element that belongs to the Grid ([toolbar](../reference/toolbar.md) in our case), you should wrap the custom markup in the [Plugin](../../../core/docs/reference/plugin.md) component to create a plugin.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+
+Create a new folder called `plugins` in the project and add a new file, `toolbar-filter.js`, with the following code to it.
```
dirk-pieterse: "the following code to it." -> the following code to it:

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+      <Plugin
+        name="ToolbarFilter"
+      >
+      </Plugin>
+    );
+  }
+}
+```
+
+The Grid's toolbar exports a [template](../../../core/docs/reference/template.md) with the `toolbarContent` name. Add the [Template](../../../core/docs/reference/template.md) component to the plugin and set its `name` attribute to `toolbarContent`. This will be the place for our custom markup.
```
dirk-pieterse: Again, why future tense?

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+        ...
+      >
+        <Template name="toolbarContent" />
+      </Plugin>
+    );
+  }
+}
+```
+
+We do not want to *replace* the default toolbar content. Rather we should *append* our content to it. To do this, add the [TemplatePlaceholder](../../../core/docs/reference/template-placeholder.md) component without parameters to the template:
```
dirk-pieterse: "We do not want to *replace* the default toolbar content. Rather we should *append* our content to it." -> We should *append* our content to default toolbar.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+        dependencies={pluginDependencies}
+      >
+        ...
+      </Plugin>
+    );
+  }
+}
+```
+
+Now, all is set to add custom markup to the `toolbarContent` template. To display a select box, add the [Select](https://material-ui.com/demos/selects/) component to the template.
```
dirk-pieterse: "all is set"?

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+          <TableHeaderRow />
+          <Toolbar />
+          <ToolbarFilter />
+        </Grid>
+      </Paper>
+    );
+  }
+}
+```
+If you run the code now, the Grid should look as follows.
```
dirk-pieterse: "look as follows." -> look as follows:

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+    );
+  }
+}
+```
+
+Now, you can filter grid rows by the `CompanyName` column.
+
+## Final Steps
+
+At the final steps, we should display the current column's caption near the select box and add the `Clear` button that clears the selected value. Since these steps do not add anything to the concepts explained earlier, we will go over them briefly:
```
dirk-pieterse: First, do you really need a section called 'Final Steps'?

dirk-pieterse: Why do you add "Since these steps do not add anything to the concepts explained earlier, we will go over them briefly"? Why future tense?

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+
+* Add the `clearToolbarFilterValue` action to the `ToolbarFilterState` plugin. The action should clear the `filterValue` in the state.
```
dirk-pieterse: This repeat the same information.

packages/dx-react-grid/docs/guides/custom-plugin-development.md
```diff
@@ -0,0 +1,709 @@
...
+  <img class="img-responsive" src="../../../../../img/guides/custom-plugin-development/add-select.png">
+</p>
+
+## Share Values Between Plugins
+
+The select box is empty now. We should populate it with values of the column by which we want to filter the grid.
+
+The `ToolbarFilter` plugin, which we implemented in the previous step, is a [UI plugin](./plugin-overview.md): it renders UI elements based on provided data. We could also enable it to modify the data or manipulate the Grid's state, but it is better to implement this functionality in a [state management plugin](./plugin-overview.md). This separation of concerns makes the plugins more flexible and maintainable.
+
+The state management plugin called `ToolbarFilterState` will prepare all necessary information for the `ToolbarFilter` plugin. Create a new `toolbar-filter-state.js` file in the `plugins` folder and add the following code to it.
```
dirk-pieterse: Future tense

PR #1753, opened 2019-01-10
refactor(react-chart): move point coordinates calculations specifics down to the pointComponent

Krijovnick:
BREAKING CHANGES:

Previously, `pointComponent` of `BarSeries` and `PieSeries` accepted precalculated fields (`width` and `d`, respectively). Now, `pointComponent` accepts fields that provide raw data for calculation. This makes `pointComponent` more flexible as the `width` and `d` fields can now be calculated the way you need.
 
The following substitutions took place:
 
- `width` => `barWidth` and `maxBarWidth`
- `d` => `innerRadius`, `outerRadius`, `maxRadius`, `startAngle`, and `endAngle`


packages/dx-react-chart/src/utils/series-helper.jsx
```diff
@@ -15,15 +15,28 @@ import {
...
     render() {
-      const { name } = this.props;
+      const {
+        name,
+        argumentField,
+        valueField,
+        scaleName,
+        seriesComponent,
+        pointComponent,
+        color,
```
DmitryBogomolov: There is no difference between `color` and `restProps` here.

packages/dx-react-chart/src/utils/series-helper.jsx
```diff
@@ -15,15 +15,28 @@ import {
...
+        ...restProps
+      } = this.props;
       const symbolName = Symbol(name);
       const seriesItem = {
         getPointTransformer,
         createHitTester,
         ...this.props,
```
DmitryBogomolov: `seriesItem` contains common series configuration - specific and appearance settings are merged later in `addSeries`.
common configuration - settings separated from `restProps` several lines above.

packages/dx-chart-core/src/plugins/series/computeds.js
```diff
@@ -166,27 +154,29 @@ const getUniqueName = (list, name) => {
...
   // *symbolName* cannot be used as it cannot be part of DOM attribute name.
   const index = series.length;
   return [...series, {
     ...props,
     name: getUniqueName(series, props.name),
     index,
-    points: createPoints(props.argumentField, props.valueField, data),
+    points: createPoints(props.argumentField, props.valueField, data, restProps),
     palette, // TODO: For Pie only. Find a better place for it.
     color: props.color || palette[index % palette.length],
```
DmitryBogomolov: `color` should be merged to `restProps` - otherwise there will be no `color` property in a series point.
`restProps` should be enumerated here instead of `color`.

Actually there is `color` property in a series point. But it is so because of `{...restProps}` in *point-collection.jsx*.

Let's add note in *point-collection.jsx* saying that `restProps` are used because of `getAnimatedStyle` and `scales`.

packages/dx-chart-core/src/plugins/series/computeds.js
```diff
@@ -81,18 +75,16 @@ export const getAreaPointTransformer = (series) => {
...
-  const width = categoryWidth * barWidth;
+  const fixedArgumentScale = fixOffset(argumentScale);
   return point => ({
     ...point,
-    x: argumentScale(point.argument) + offset,
+    x: fixedArgumentScale(point.argument),
     y: valueScale(point.value),
     y1,
-    width,
+    spacingForBar: getWidth(argumentScale),
```
DmitryBogomolov: We should verify this name.

PR #1764, opened 2019-01-15
chore(react-core): fix sizer crash on 'scroll' event after unmount

DmitryBogomolov:


packages/dx-react-core/src/sizer.test.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
+import { Sizer } from './sizer';
+
+describe('Sizer', () => {
+  const divProto = (document.createElement('div') as HTMLDivElement).constructor.prototype;
```
ushkal: Can we use a mock object as a Container instead? 
{ addEventListener: jest.fn(), ... }

DmitryBogomolov: `Container` must be a React component instance - plain object cannot be used instead.

The other option is to wrap `document.createElement` and add mock `addEventListener` and `removeEventListener` to every created *div*.
Overriding those methods in *div* prototype seems just simpler.

PR #1782, opened 2019-01-18
chore(react-chart): fix colors in pie series

Krijovnick:


packages/dx-chart-core/src/plugins/series/computeds.test.js
```diff
@@ -481,7 +479,43 @@ describe('addSeries', () => {
           },
         ],
         color: 'c2',
-        palette,
+      },
+    ]);
+  });
+
+  it('should create points for pie', () => {
```
DmitryBogomolov: There is no *pie* specific here - it should be something like *'should calculate point color'* or *'should calculate point color with "getPointColor" function'*.

packages/dx-chart-core/src/plugins/series/computeds.test.js
```diff
@@ -481,7 +479,43 @@ describe('addSeries', () => {
...
+          {
+            argument: 'b', value: 2, index: 1, userOptions: 'userOptions', color: 'point color',
+          },
+          {
+            argument: 'd', value: 4, index: 3, userOptions: 'userOptions', color: 'point color',
+          },
+        ],
+        color: 'c2',
       },
     ]);
```
DmitryBogomolov: Check that `getPointColor` has been called several times. Something like that:
```js
expect(getPointColor.mock.calls).toEqual([
  [palette, 0], [palette, 1], [palette, 2],
]);
```

PR #1791, opened 2019-01-22
chore: allow to measure demo rendering time

DmitryBogomolov:


packages/dx-demo-shell/src/demo-viewer/perf-wrapper.jsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  WrappedDemo.propTypes = {
+    // False alarm - used in `getDerivedStateFromProps`.
+    count: PropTypes.number, // eslint-disable-line react/no-unused-prop-types
+  };
+
+  WrappedDemo.defaultProps = {
+    count: 16,
```
Krijovnick: Maybe, the `count` property make changeable?  For example, I create demo with 10 thousand points (it is not very large number) and I want to measure the performance. But in the browser I can see that page unresponsive. When I change that property to 5 - it is all OK.

DmitryBogomolov: Done - `<path-to-demo>/perf/5`.

PR #1793, opened 2019-01-22
fix(react-grid): disallow rendering a checkbox in incorrect column when grouping is enabled

ushkal:


packages/dx-react-grid/src/plugins/table-group-row.jsx
```diff
@@ -17,6 +17,7 @@ const pluginDependencies = [
   { name: 'GroupingState' },
   { name: 'Table' },
   { name: 'DataTypeProvider', optional: true },
+  { name: 'TableSelection', optional: true },
```
SergeyAlexeev: What about [docs](https://devexpress.github.io/devextreme-reactive/react/grid/docs/reference/table-group-row/)?

PR #1803, opened 2019-01-28
fix(react-chart): fix-event-handlers-in-safari

DmitryBogomolov:
Handles *mousemove* and *mouseleave* events if either *pointer* or *touch* events are not available.

### DIFF
--- a/packages/dx-chart-core/src/utils/event-tracker.test.js
+++ b/packages/dx-chart-core/src/utils/event-tracker.test.js
@@ -165,7 +165,7 @@ describe('EventTracker', () => {
       });
     });
 
-    it('should create only pointer move handlers', () => {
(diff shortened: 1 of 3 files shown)


packages/dx-chart-core/src/utils/event-tracker.test.js
```diff
@@ -171,16 +171,16 @@ describe('EventTracker', () => {
       });
 
       expect(handlers).toEqual({
```
Krijovnick: In that test we check mouse move handlers, thus the name of the test should not be `...pointer move handlers`.

DmitryBogomolov: Removed *pointer* occurrences.

PR #1812, opened 2019-01-31
feat(dx-react-grid): rtl support for TableColumnResizing

yairtawil:
example:

```jsx
       <Grid rows={rows} columns={columns} isRtl={true}>
          <DragDropProvider />
          <Table />
          <TableColumnResizing
            columnWidths={columnWidths}
            onColumnWidthsChange={this.changeColumnWidths}
          />
          <TableColumnReordering defaultOrder={["city", "sex", "name"]} />
          <TableHeaderRow />
        </Grid>
```

### DIFF
--- a/packages/dx-react-grid/src/plugins/table-header-row.jsx
+++ b/packages/dx-react-grid/src/plugins/table-header-row.jsx
@@ -13,6 +13,7 @@ import {
   TABLE_HEADING_TYPE,
 } from '@devexpress/dx-grid-core';
 
+const isRtlComputed = (({ isRtl }) => isRtl);
 const tableHeaderRowsComputed = ({ tableHeaderRows }) => tableRowsWithHeading(tableHeaderRows);
 
 export class TableHeaderRow extends React.PureComponent {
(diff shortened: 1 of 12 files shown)


yotamberk:
+1

packages/dx-react-grid/src/plugins/table-header-row.jsx
```diff
@@ -41,6 +42,7 @@ export class TableHeaderRow extends React.PureComponent {
           { name: 'TableColumnResizing', optional: true },
         ]}
       >
+        <Getter name="isRtl" computed={isRtlComputed} />
```
SergeyAlexeev: It isn't required to define such Getter in each plugin. You can use the previously calculated value in TemplateConnector directly.

SergeyAlexeev:
Hi @yairtawil,

Thank you for your great work! I'd like to add that the 'RTL Support' should affect all Grid plugins to keep UI and behavior consistent. Do you have plans to process other plugins as well?
 
As for the pull request, there aren't any [tests](https://github.com/DevExpress/devextreme-reactive/blob/master/README_DEVELOPERS.md#tests-and-ci). We need to be sure that all the functionality works as expected.
 
Moreover, each feature should contain documentation and demos. Otherwise, other developers wont' be able to use our components without additional support.

PR #1816, opened 2019-01-31
refactor(chart-core): migrate chart core to TS

Krijovnick:


packages/dx-chart-core/package.json
```diff
@@ -35,25 +35,22 @@
...
-    "babel-jest": "^23.6.0",
     "core-js": "^2.6.1",
-    "eslint": "^5.11.0",
-    "eslint-config-airbnb-base": "^13.1.0",
-    "eslint-plugin-filenames": "^1.3.2",
-    "eslint-plugin-import": "^2.14.0",
-    "eslint-plugin-jest": "^22.1.2",
     "jest": "^23.6.0",
     "rollup": "^0.68.2",
     "rollup-plugin-babel": "^4.1.0",
```
ushkal: We should remove "rollup-plugin-babel" too

packages/dx-chart-core/src/plugins/scale/computeds.ts
```diff
@@ -2,39 +2,51 @@ import { extent } from 'd3-array';
...
+const mergeContinuousDomains = (
+  domain, items,
+) => extent([...domain, ...items]);
+const mergeDiscreteDomains = (
+  domain, items,
+) => Array.from(new Set([...domain, ...items]));
 
-const getArgument = point => point.argument;
-const getValue = point => point.value;
+const getArgument = (point): Point => point.argument;
```
ushkal: Should be (point: Point) ?

packages/dx-chart-core/src/plugins/scale/computeds.ts
```diff
@@ -2,39 +2,51 @@ import { extent } from 'd3-array';
...
+const getValue = (point): Point => point.value;
 
 const extendDomain = (target, items) => {
   const merge = target.isDiscrete ? mergeDiscreteDomains : mergeContinuousDomains;
   Object.assign(target, { domain: merge(target.domain, items) });
 };
 
-const calculateDomains = (domains, seriesList) => {
+const calculateDomains: PureComputed<[any, Series[]]> = (domains, seriesList) => {
```
ushkal: Type can be specified here: any -> Domain[]

packages/dx-chart-core/src/utils/scale.ts
```diff
@@ -0,0 +1,19 @@
...
+} from '../types';
+
+export const isHorizontal: PureComputed<[string], boolean> = name => name === ARGUMENT_DOMAIN;
+
+export const getWidth: PureComputed<[Scale], number> = scale => (
+  scale.bandwidth ? scale.bandwidth() : 0
+);
+
+// tslint:disable-next-line: ter-arrow-parens
+export const getValueDomainName: PureComputed<[string]> = (name?) => name || VALUE_DOMAIN;
```
ushkal: Should be nullable - "string?"

Krijovnick: yes, i'll fix it

packages/dx-chart-core/src/types/utils.types.ts
```diff
@@ -0,0 +1,53 @@
+import { PureComputed, CustomFunction } from '@devexpress/dx-core';
+import { Point, Series, Target, BarPoint, PiePoint, ScatterPoint } from './chart-core.types';
+
+export  type MakePath = PureComputed<[], any>;
+export type List = {
```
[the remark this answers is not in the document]
Krijovnick: I think the name `PointDistance` will be more clearer because the object characterizes the distance for each point.

packages/dx-chart-core/src/plugins/axis/computeds.ts
```diff
@@ -41,14 +46,40 @@ const createVerticalOptions = (position, tickSize, indentFromAxis) => {
...
+  readonly y2: number;
+  readonly xText: number;
+  readonly yText: number;
+  readonly dominantBaseline: string;
+  readonly textAnchor: string;
+  readonly text: string;
+};
+type AxisCoordinatesResult = {
+  readonly ticks: ReadonlyArray<Tick>;
+  readonly sides: Readonly<[number, number]>;
```
ushkal: These readonly modifiers can be removed since PureComputed makes argument types deep readonly

packages/dx-chart-core/src/plugins/axis/computeds.ts
```diff
@@ -41,14 +46,40 @@ const createVerticalOptions = (position, tickSize, indentFromAxis) => {
   };
 };
 
-export const axisCoordinates = ({
+type AxisCoordinatesArg = {
```
ushkal: I suggest we place all custom defined types and interfaces in separate 'types.ts' files - thus a code will be a bit easier to read and if we want to see a type definition we can use intelliSense/F12

Krijovnick: done

PR #1824, opened 2019-02-05
fix(react-grid): prevent converting group row value to a string if it is undefined

### LINKED ISSUE #1761: React Grid grouping row value is `undefined`

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour
When grouping by a column that could be undefined, the resulted row for undefined group looks like this:

```js
{
    compoundKey: "undefined",
    groupedBy: "column-name",
    key: "undefined",
    value: "undefined",
    ...
}
```

I understand that the key should be string, but don't get why value should be "undefined" as well.

After trying to see why this is happening, i found the next piece of code in integrated-grouping plugin that explains why value is "undefined":

```code
const { key, value = key } = groupCriteria(getCellValue(row, columnName), row);
```

Are there any reasons for having this behavior?

### Environment

- devextreme-reactive: 1.10.0

ushkal:
fixes #1761


packages/dx-grid-core/src/plugins/integrated-grouping/computeds.ts
```diff
@@ -29,11 +29,13 @@ export const groupedRows: GroupedRowsFn = (
...
   const groups = rows
     .reduce((acc, row) => {
-      const { key, value } = groupCriteria(getCellValue(row, columnName), row);
+      const rawValue = getCellValue(row, columnName);
+      const { key, value } = groupCriteria(rawValue, row);
       const sameKeyItems = acc.get(key);
 
       if (!sameKeyItems) {
-        acc.set(key, [value || key, key, [row]]);
+        const groupingValue = value === rawValue ? value : value || key;
```
SergeyAlexeev: Just a question: is there any case when value !== rawValue? I've tried to use `undefined`, `null` as a column value. `value === rawValue` - is always `true`.

ushkal: Yes - when custom columnCriteria is specified and it returns only a key, thus a value is undefined but rawValue may be not

PR #1828, opened 2019-02-06
fix(react-core): notify Sizer content when container size is increasing

### LINKED ISSUE #1763: VirtualTable not rendering all rows when height changes

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

On initial load, rows=[] and height is 160 (the height of the no data component). Rows are then fetched from the API, and we recalculate the height and pass it to the virtual table. The height is correct in the app, but only 6 rows are rendered, and the rest of the height is filled with some dummy row.

If I return null instead when the rows array is empty, the first render pass is correct, so this looks like a similar issue to https://github.com/DevExpress/devextreme-reactive/issues/1627 - maybe it was not properly fixed in the most recent release.

### Expected Behaviour

The table should render all rows passed in to it.

https://codesandbox.io/s/4jwj2kovx9

Click add more rows - note that even though height is being calculated correctly, only 6 rows appear in the table and in the DOM

### Environment

- devextreme-reactive: 1.10.0
- react: 16.6.3
- browser: Chrome
- bootstrap: none
- react-bootstrap: none
- material-ui: 3.7.0

ushkal:
fixes #1763

### DIFF
--- a/packages/dx-react-core/src/sizer.tsx
+++ b/packages/dx-react-core/src/sizer.tsx
@@ -5,6 +5,9 @@ import { findDOMNode } from 'react-dom';
 import { RefHolder } from './ref-holder';
 
 const styles = {
+  root: {
+    position: 'relative',
+  },
   triggersRoot: {
     position: 'absolute',
     top: 0,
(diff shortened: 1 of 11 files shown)


nthgol:
so this is coming out in 1.10.2?

PR #1838, opened 2019-02-11
chore(react-chart): use reference element in tooltip

DmitryBogomolov:
Replaces real svg element with a [reference object](https://popper.js.org/popper-documentation.html#referenceObject) instance.


packages/dx-chart-core/src/plugins/tooltip/computeds.ts
```diff
@@ -12,6 +14,28 @@ export const getParameters = (series: SeriesList, target: Target): TooltipParame
   };
 };
 
+export const createReference = (
+  rect: Rect, rootRef: React.RefObject<Element>,
+): TooltipReference => ({
+  // These two fields together with *width* and *height* are left with stub data for
+  // simplicity reasons - they seem to be unused by *Popper*.
```
ushkal: height and width are used to calculate arrow position - https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/modifiers/arrow.js#L44

DmitryBogomolov: Looks like they are not.
bs4-tooltip-arrow

It seems that they are used to address fields not from a reference object but from *data.offsets* option.

https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/modifiers/arrow.js#L69
https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/modifiers/arrow.js#L41
https://popper.js.org/popper-documentation.html#dataObject

PR #1849, opened 2019-02-14
fix(react-grid): fix borders of fixed columns in Edge

ushkal:


packages/dx-react-grid-bootstrap3/src/templates/table-fixed-cell.test.jsx
```diff
@@ -0,0 +1,93 @@
...
+    });
+  });
+
+  it('should pass custom styles to the root element', () => {
+    const tree = shallow((
+      <FixedCell {...defaultProps} style={{ color: 'white' }} />
+    ));
+
+    expect(tree.prop('style')).toMatchObject({
+      color: 'white',
```
SergeyAlexeev: I'd check `{ color: 'white' }` doesn't completely override styles but extends them. E.g. check that the `position: sticky` exists too.

PR #1856, opened 2019-02-21
chore(docs): revise headers

RomanTsukanov:


packages/dx-react-common/docs/guides/custom-theme.md
```diff
@@ -14,7 +14,7 @@ The theme consists of two directories: `plugins` (plugins that we customize) and
 ### Components
 As we mentioned earlier, plugins render UI elements via render props. All render props have the postfix `Component`. For instance, the [Table](https://devexpress.github.io/devextreme-reactive/react/grid/docs/reference/table/) plugin contains the `tableComponent`, `headComponent`, `cellComponent`, and others.
 
-#### Handle the Props
+### Handle the Props
```
ushkal: This section belongs to the 'Components' thus should be indented

ushkal:
LGTM

packages/dx-react-scheduler/docs/guides/fundamentals.md
```diff
@@ -23,7 +23,7 @@ To specify the date that should be initially displayed in the date navigator and
 
 .embedded-demo({ "path": "scheduler-basic/basic-setup", "showThemeSelector": true })
 
-## Appearance Customization
+## Customize the Appearance
```
MaximKudriavtsev: We very often add this link to our customers in issues. Can we add a redirect to this new ref?

MaximKudriavtsev: But in the Fundamentals section, we have only one big subsection) In this case redirect is not need

RomanTsukanov: AFAIK, there's no way to redirect from an anchor link, so I'm afraid we can't do this

PR #1866, opened 2019-02-26
fix(react-grid): fix virtual table column width in Edge

ushkal:


packages/dx-react-grid/src/components/table-layout/__snapshots__/virtual-table-layout.test.tsx.snap
```diff
@@ -30,40 +30,35 @@ exports[`VirtualTableLayout should render correct layout 1`] = `
                 "name": "a",
               },
               "key": "a",
-              "preferMinWidth": true,
-              "width": 120,
+              "width": 160,
```
ushkal: default table width is 800 and here's 5 columns

PR #1879, opened 2019-03-01
refactor(react-chart): move domains Getter to series component

DmitryBogomolov:
The **domains** *Getter* is moved from `ChartCore` to series component.
The behavior is changed as well. Previously all domains were built at once. Now domains are updated every time a series is added.

### DIFF
--- a/packages/dx-chart-core/src/plugins/scale/computeds.ts
+++ b/packages/dx-chart-core/src/plugins/scale/computeds.ts
@@ -11,6 +11,7 @@ import {
 const makeDomain = ({ factory, modifyDomain }: DomainOptions): DomainInfo => ({
   domain: [],
   factory,
+  isDiscrete: !!(factory && isDiscrete(factory)),
   modifyDomain,
 });
 
@@ -44,7 +45,7 @@ const getArgument: GetItemFn = point => point.argument;
 const getValue: GetItemFn = point => point.value;
 
 /** @internal */
-export const scaleLinear: FactoryFn = d3ScaleLinear;
+export const scaleLinear: FactoryFn = d3ScaleLinear as any;
 /** @internal */
 export const scaleBand: FactoryFn = () => (
   d3ScaleBand().paddingInner(0.3).paddingOuter(0.15) as any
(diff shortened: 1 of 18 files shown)


packages/dx-chart-core/src/plugins/scale/computeds.ts
```diff
@@ -3,125 +3,122 @@ import { scaleLinear as d3ScaleLinear, scaleBand as d3ScaleBand } from 'd3-scale
...
-export const computeDomains: ComputeDomainsFn = (domains, seriesList) => {
-  const result = copy(domains);
-  collectDomainsFromSeries(result, seriesList);
-  calculateDomains(result, seriesList);
-  customizeDomains(result);
-  return result;
+export const extendDomains: ExtendDomainsFn = (domains, series) => {
+  const argumentDomain = updateDomain(
+    domains[ARGUMENT_DOMAIN], series, getArgument, getArgumentDomainItems);
+  const valueDomainName = getSeriesValueDomainName(series);
```
Krijovnick: Why is `getValueDomainName` not called directly here?

packages/dx-chart-core/src/plugins/scale/computeds.ts
```diff
@@ -3,125 +3,122 @@ import { scaleLinear as d3ScaleLinear, scaleBand as d3ScaleBand } from 'd3-scale
...
+  const valueDomain = updateDomain(
```
Krijovnick: Why is the `updateDomainItems` method not called directly here?

packages/dx-chart-core/src/plugins/scale/computeds.ts
DmitryBogomolov: `getSeriesValueDomainName` is removed.

packages/dx-chart-core/src/plugins/scale/computeds.ts
DmitryBogomolov: `updateDomain` encloses and orders smaller domains updates (*factory* -> *items*) and is applied similarly to argument and value domains.

PR #1894, opened 2019-03-10
chore(react-chart): fix stacks processing after "domains" Getter movement

DmitryBogomolov:
`getPointTransformer.getValueDomains` is removed.

### DIFF
--- a/packages/dx-react-chart/src/plugins/stack.tsx
+++ b/packages/dx-react-chart/src/plugins/stack.tsx
@@ -1,11 +1,13 @@
 import * as React from 'react';
 import { Plugin, Getter, Getters } from '@devexpress/dx-react-core';
-import { getStackedSeries } from '@devexpress/dx-chart-core';
+import { getStackedSeries, getStackedDomains } from '@devexpress/dx-chart-core';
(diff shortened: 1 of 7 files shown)


packages/dx-react-chart/src/plugins/stack.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { Plugin, Getter, Getters } from '@devexpress/dx-react-core';
-import { getStackedSeries } from '@devexpress/dx-chart-core';
+import { getStackedSeries, getStackedDomains } from '@devexpress/dx-chart-core';
 import { StackProps, StacksOptions, OffsetFn, OrderFn } from '../types';
 import {
   stackOrderNone,
   stackOffsetDiverging,
 } from 'd3-shape';
 
+const doGetStackedDomains = ({ domains, series }: Getters) => getStackedDomains(domains, series);
```
Krijovnick: I think this method name is strange. Can we use another one? For example, 'getDomains'?

DmitryBogomolov: Renamed.

PR #1895, opened 2019-03-11
chore(react-chart): add some more content to "series interaction" demo

DmitryBogomolov:
series-interaction-demo


packages/dx-react-chart-demos/src/demo-sources/bar-chart/series-interaction.jsxt
```diff
@@ -0,0 +1,202 @@
...
+const tooltipContentBodyStyle = {
+  paddingTop: 0,
+};
+const formatTooltip = d3Format.format(',.2r');
+const TooltipContent = (props) => {
+  const { targetItem, text, ...restProps } = props;
+  return (
+    <div>
+      <div>
+        <Tooltip.Content
```
Krijovnick: Tooltip.Content is undocumented. Maybe we should do that?

DmitryBogomolov: It is already partially [documented](https://devexpress.github.io/devextreme-reactive/react/chart/docs/reference/tooltip/#tooltip.contentprops). Let's fix it in a separate PR.

Krijovnick: ok

PR #1908, opened 2019-03-14
refactor(react-grid): add an ability to pass immutable data

MaximKudriavtsev:


packages/dx-grid-core/package.json
```diff
@@ -49,7 +49,8 @@
     "tslint": "^5.13.0",
     "tslint-config-airbnb": "^5.11.1",
     "typescript": "^3.3.3333",
-    "typescript-tslint-plugin": "^0.3.1"
+    "typescript-tslint-plugin": "^0.3.1",
+    "seamless-immutable": "^7.1.4"
```
DmitryBogomolov: Is one more dependency (even "dev") actually required?
It seems that the following snippet would cover the case.
```js
const makeImmutable = (value) => {
  const target = value.slice();
  target.slice = () => makeImmutable(value);
  target.sort = () => {
    throw new Error('Attempt to modify!');
  };
  return target;
};
```

packages/dx-grid-core/src/plugins/custom-tree-data/computeds.ts
```diff
@@ -70,7 +70,7 @@ export const expandedTreeRows: ExpandedTreeRowsFn = (
 
   let currentExpanded = true;
   let currentLevel = 0;
-  return rows.reduce((acc, row) => {
+  return Array.prototype.slice.call(rows).reduce((acc, row) => {
```
DmitryBogomolov: Is it required?

packages/dx-grid-core/src/plugins/integrated-grouping/computeds.ts
```diff
@@ -27,7 +27,7 @@ export const groupedRows: GroupedRowsFn = (
   const { columnName } = grouping[0];
   const groupCriteria = (getColumnCriteria && getColumnCriteria(columnName))
     || defaultColumnCriteria;
-  const groups = rows
+  const groups = Array.prototype.slice.call(rows)
```
DmitryBogomolov: Is it required?

packages/dx-grid-core/src/plugins/sorting-state/reducers.ts
```diff
@@ -9,10 +9,10 @@ export const changeColumnSorting: PureReducer<ColumnSortingState, ChangeSortingP
 
   let nextSorting: any[] = [];
   if (keepOther === true) {
-    nextSorting = sorting.slice();
+    nextSorting = Array.prototype.slice.call(sorting);
   }
   if (Array.isArray(keepOther)) {
-    nextSorting = sorting.slice()
+    nextSorting = Array.prototype.slice.call(sorting)
```
DmitryBogomolov: `slice` is not required because `filter` creates a copy.

packages/dx-grid-core/src/utils/merge-sort.ts
```diff
@@ -57,8 +57,8 @@ export default (
     return 0;
   },
 ) => {
-  const result = array.slice();
-  const auxiliary = array.slice();
+  const result = Array.prototype.slice.call(array);
```
DmitryBogomolov: From the code of `merge` it seems that it does not modify `result`, isn't it?

packages/dx-react-grid/src/components/table-layout.tsx
```diff
@@ -65,7 +65,7 @@ class TableLayoutBase extends React.PureComponent<TableLayoutCoreProps, TableLay
 
     const isFixedWidth = columns.filter(column => column.width === undefined).length === 0;
     if (isFixedWidth) {
-      result = result.slice();
+      result = Array.prototype.slice.call(result);
```
DmitryBogomolov: May be just like that?

```js
result = [...result, { ... }];
```

packages/dx-grid-core/src/plugins/custom-tree-data/computeds.ts
MaximKudriavtsev: Yeah, `seamless-immutable` lib. replace all basic array methods. I have created tests to show it needs.

packages/dx-grid-core/src/plugins/integrated-grouping/computeds.ts
MaximKudriavtsev: Without these changes, a demo will break down.

packages/dx-grid-core/src/plugins/sorting-state/reducers.ts
MaximKudriavtsev: Without these changes, a demo will break down. You can see it if you remove my changes and run tests

packages/dx-grid-core/src/utils/merge-sort.ts
MaximKudriavtsev: Tests display other, or I don't understand something?

packages/dx-react-grid/src/components/table-layout.tsx
MaximKudriavtsev: All tests pass. Thank you

packages/dx-grid-core/src/utils/merge-sort.ts
DmitryBogomolov: My mistake - the exported function itself modifies the `array`. So copy is required.

packages/dx-grid-core/package.json
MaximKudriavtsev: In our documentation, we use the `seamless-immutable` lib and say to customers use it in your projects. I think we should test this lib without any mocks. But on the other side, we should pass any other `immutable` libraries into our controls.

MaximKudriavtsev: Because in the `seamless-immutable` the `reduce` method can modify the array, too. I don't know how, but tests display it.

DmitryBogomolov: How does it modify array?
https://github.com/rtfeldman/seamless-immutable/blob/master/src/seamless-immutable.js#L91
https://github.com/rtfeldman/seamless-immutable/blob/master/src/seamless-immutable.js#L132

packages/dx-grid-core/src/plugins/custom-tree-data/computeds.test.ts
```diff
@@ -255,7 +257,52 @@ describe('CustomTreeData Plugin computeds', () => {
...
       };
       const getRowId = row => rows.indexOf(row);
-      const expandedRowIds = [1];
+
+      const expandedLinearizedRows = {
+        /* tslint:disable ter-indent align */
+        rows: [
+          { a: 0 },
+          { a: 1 },
+            { a: 1, b: 1 },
```
SergeyAlexeev: This indent looks strange.

packages/dx-grid-core/src/plugins/table-edit-row/computeds.ts
```diff
@@ -27,7 +27,7 @@ export const tableRowsWithEditing: TableRowsWithEditingFn = (
     }));
 
   return [
-    ...addedTableRows.reverse(),
+    ...Array.prototype.slice.call(addedTableRows).reverse(),
```
SergeyAlexeev: Can we create an util in dx-core, for instance, that will look like below:

```js
const slice = arr => Array.prototype.slice.call(arr); // slice can be renamed to copy as well

// usage
...slice(addedTableRows).reverse();
```

The reason: 'Array.prototype.slice' looks too verbose.

packages/dx-react-grid/src/components/table-layout.test.tsx
```diff
@@ -231,6 +232,19 @@ describe('TableLayout', () => {
         expect(getAnimations)
           .toHaveBeenLastCalledWith(expect.anything(), expect.anything(), 400, new Map());
       });
+
+      it('should work with immutable state', () => {
```
SergeyAlexeev: You are using the 'state' word in a test name. Actually, you are using immutable array as a property.

packages/dx-grid-core/src/plugins/grouping-state/reducers.ts
```diff
@@ -8,7 +8,7 @@ import {
 const applyColumnGrouping: PureReducer<Grouping[], ChangeGroupingPayload> = (
   grouping, { columnName, groupIndex },
 ) => {
-  const nextGrouping = grouping.slice();
+  const nextGrouping = slice(grouping);
```
DmitryBogomolov: If none of the following conditions is true then the function result is falsely different - `grouping` is not changed actually but instance reference is changed.

packages/dx-grid-core/src/plugins/integrated-sorting/computeds.ts
```diff
@@ -22,7 +22,7 @@ const defaultCompare = (a: any, b: any) => {
 
 const createCompare: CreateCompareFn = (
   sorting, getColumnCompare, getComparableValue,
-) => sorting.slice()
+) => slice(sorting)
```
DmitryBogomolov: Why not `reduceRight`?

packages/dx-grid-core/src/plugins/sorting-state/reducers.ts
```diff
@@ -9,10 +9,10 @@ export const changeColumnSorting: PureReducer<ColumnSortingState, ChangeSortingP
 
   let nextSorting: any[] = [];
   if (keepOther === true) {
-    nextSorting = sorting.slice();
+    nextSorting = slice(sorting);
   }
   if (Array.isArray(keepOther)) {
-    nextSorting = sorting.slice()
+    nextSorting = slice(sorting)
```
DmitryBogomolov: `slice` is not required here (`filter` creates new list)

packages/dx-grid-core/src/plugins/sorting-state/reducers.ts
```diff
@@ -9,10 +9,10 @@ export const changeColumnSorting: PureReducer<ColumnSortingState, ChangeSortingP
 
   let nextSorting: any[] = [];
   if (keepOther === true) {
```
DmitryBogomolov: If `keepOther` is `true` and neither `columnSortingIndex > -1` nor `direction !== null` are `true` then function result is result is falsely different.

packages/dx-grid-core/src/plugins/table-band-header/helpers.ts
```diff
@@ -16,11 +17,9 @@ export const isNoDataColumn = (columnType: symbol) => columnType !== TABLE_DATA_
 export const getColumnMeta: GetColumnBandMetaFn = (
   columnName, bands, tableRowLevel,
   level = 0, title = null, result = null,
-) => bands.reduce((acc, column) => {
+) => slice(bands).reduce((acc, column) => {
```
DmitryBogomolov: Why is `slice` used here?

SergeyAlexeev:
Common question: what about Chart and Scheduler? Are we going to process them in separate PRs?

MaximKudriavtsev:
> Common question: what about Chart and Scheduler? Are we going to process them in separate PRs?

@SergeyAlexeev I think yes, owners of these products `chart` and `scheduler` understand own code better.

packages/dx-grid-core/src/plugins/sorting-state/reducers.ts
MaximKudriavtsev: There is `slice` in the next conditions

PR #1921, opened 2019-03-19
refactor(react-scheduler): change drag and drop functions

MaximKudriavtsev:


packages/dx-react-scheduler/src/plugins/drag-drop-provider.jsx
```diff
@@ -162,35 +143,22 @@ export class DragDropProvider extends React.PureComponent {
       if (targetType === VERTICAL_TYPE) {
```
ushkal: It seems that the code computing this.allDayRects and this.timeTableRects can be extracted to separate function too. It will require 5-6 parameters but calculateBoundaries will be easier to read

packages/dx-react-scheduler/src/plugins/drag-drop-provider.jsx
```diff
@@ -162,35 +143,22 @@ export class DragDropProvider extends React.PureComponent {
...
         this.timeTableRects = verticalTimeTableRects(
           draftAppointments, startViewDate, endViewDate,
-          excludedDays, viewCellsData, cellDuration, timeTableCells,
+          excludedDays, viewCellsData, cellDurationMinutes, timeTableCells,
         );
       } else {
         this.timeTableRects = horizontalTimeTableRects(
```
ushkal: timeTableRects and allDayRects are better to rename to reflect their meanings - draft appointment geometry information and metadata

packages/dx-react-scheduler/src/plugins/drag-drop-provider.jsx
```diff
@@ -226,26 +178,30 @@ export class DragDropProvider extends React.PureComponent {
...
+                  onChange={this.onPayloadChange({ commitChangedAppointment, stopEditAppointment })}
                 >
-                  <TemplatePlaceholder />
-                </DropTarget>
-              </DragDropProviderCore>
-            )}
+                  <DropTarget
+                    onOver={calculateBoundariesByMove}
+                    onEnter={calculateBoundariesByMove}
+                    onDrop={this.onDrop({ commitChangedAppointment, stopEditAppointment })}
```
SergeyAlexeev: Can we call `handleDrop` directly? `onDrop` just returns a wrapper around `handleDrop`.

MaximKudriavtsev: No, the `onDrop` property required a function and call it with argument `{ payload, clientOffset }`. In this case, we should pass our actions to this function

PR #1945, opened 2019-04-02
chore(react-chart): fix default tick format for small viewport

DmitryBogomolov:


Krijovnick:
The title of the PR should be without hyphen between the words.

PR #1952, opened 2019-04-08
Fix drag and drop transform for IE11

ssg-luke:
Fixes drag and drop transform on IE11. Otherwise element being dragged is stuck in top left corner


MaximKudriavtsev:
The DevExtreme React Grid available for 3 themes `material-ui`, `bootstrap4` and `bootstrap3`. This PR is not contained changes for `bootstrap3/4`.

There aren't any [tests](https://github.com/DevExpress/devextreme-reactive/blob/master/README_DEVELOPERS.md#tests-and-ci). We need to be sure that all the functionality works as expected.

PR #1964, opened 2019-04-11
fix(react-grid-bootstrap3): add fallback position sticky css property in Safari

churkin:


packages/dx-react-grid-bootstrap3/src/templates/table-fixed-cell.jsx
```diff
@@ -1,8 +1,21 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      stickyPosition: getStickyPosition(),
+    };
+  }
+
+  componentDidMount() {
+    this.setState({ stickyPosition: getStickyPosition() });
```
ushkal: We need to avoid re-render here when stickyPosition hasn't changed - like in Table component

churkin: @ushkal  It is a Pure component

packages/dx-react-grid-bootstrap3/src/templates/table-container.jsx
```diff
@@ -16,6 +16,8 @@ export const TableContainer = ({
       WebkitOverflowScrolling: 'touch',
       border: 0,
       margin: 0,
+      // NOTE: fix sticky positioning in Safari
```
MaximKudriavtsev: You can test these styles and delete this comment

churkin: https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-grid-material-ui/src/templates/table-container.jsx#L11

PR #1974, opened 2019-04-15
chore(react-chart): fix animation for scatter

Krijovnick:


packages/dx-react-chart-demos/src/demo-sources/line-chart/bootstrap4/step-line-with-gaps.jsx
```diff
@@ -12,24 +12,16 @@ import { Animation } from '@devexpress/dx-react-chart';
...
+  const { value } = props;
   if (value) {
     return (
-      <path
-        fill={color}
-        transform={`translate(${x} ${y})`}
-        d={symbol().size([10 ** 2]).type(symbolCircle)()}
-        style={style}
-      />
+      <ScatterSeries.Point {...props} {...pointOptions} />
```
DmitryBogomolov: Why is there no change for material part?

Krijovnick: Done

PR #1975, opened 2019-04-16
chore(react-core): fix TemplatePlaceholder updating

DmitryBogomolov:
## What

Given a plugin

```jsx
const Tester = ({ data }) => (
  <Plugin name="Tester">
    <Template name="test">
      <TemplatePlaceholder />
      <div>{data}</div>
    </Template>
  </Plugin>
);
```

The content is changed from

```jsx
<Tester data={1} />
<Tester data={2} />
<Tester data={3} />
```

to

```jsx
<Tester data={1} />
<Tester data={2} />
```

The resulting markup is expected to change from

```jsx
<div>1</div>
<div>2</div>
<div>3</div>
```

to

```jsx
<div>1</div>
<div>2</div>
```

Though it actually is

```jsx
<div>1</div>
<div>2</div>
<div>2</div>
```

## Why

When *Tester (3)* is removed the *previous* state

```jsx
<TemplatePlaceholder /> (from 3)
<div>3</div>
```

is compared to the *next* state

```jsx
<TemplatePlaceholder /> (from 2)
<div>2</div>
```

As a result, *TemplatePlaceholder* element is taken from the *previous* state (because it is considered as **not changed**) and *div* is updated (content is changed from *3* to *2*).
*TemplatePlaceholder (from 3)* then translates to

```jsx
<div>1</div>
<div>2</div>
```

(because it is *TemplatePlaceholder* from `<Tester data={3} />`).
And so we have

```jsx
<div>1</div>
<div>2</div>
<div>2</div> (updated from 3)
```

If it was *TemplatePlaceholder (from 2)* then it would translate to

```jsx
<div>1</div>
```

and result would be

```jsx
<div>1</div>
<div>2</div> (updated from 3)
```

## Expected

*TemplatePlaceholder* should take into account the template it belongs to.
It already has custom `shouldComponentUpdate` - method should be updated.

### DIFF
--- a/packages/dx-react-core/src/plugin-based/template-placeholder.test.tsx
+++ b/packages/dx-react-core/src/plugin-based/template-placeholder.test.tsx
@@ -322,4 +322,40 @@ describe('TemplatePlaceholder', () => {
     expect(() => { wrapper.unmount(); })
       .not.toThrow();
   });
+
+  it('should return content of the associated template', () => {
+    const Tester = ({ name }) => {
+      return (
+        <Plugin name="Tester">
+          <Template name="test">
+            <TemplatePlaceholder />
+            <div className={name}>{name}</div>
+          </Template>
+        </Plugin>
+      );
+    };
+
+    const Root = ({ enabled }) => (
+      <div className="container">
+        <PluginHost>
+          <Template name="root">
+            <TemplatePlaceholder  name="test" />
+          </Template>
+          <Template name="test">
+            <div className="root">root</div>
+          </Template>
+
+          <Tester name="t1" />
+          {enabled && <Tester name="t2" />}
+        </PluginHost>
+      </div>
+    );
+
+    const tree = mount(<Root enabled={true} />);
+    tree.setProps({ enabled: false });
+
+    expect(tree.find('.container').html()).toEqual((
+      '<div class="container"><div class="root">root</div><div class="t1">t1</div></div>'
+    ));
+  });
 });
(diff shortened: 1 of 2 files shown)


DmitryBogomolov:
There is a side effect.

Given a pair of plugins using same template but with different content

```jsx
<Plugin name="Tester">
  <Template name="test">
    <TemplatePlaceholder />
    <div>{data}</div>
  </Template>
</Plugin>

<Plugin name="Target">
  <Template name="test">
    <TemplatePlaceholder />
    { some other content }
  </Template>  
</Plugin>
```

----
When this

```jsx
<Tester data={1} />
<Tester data={2} />
<Tester data={3} />
```

changes to

```jsx
<Tester data={1} />
<Tester data={2} />
<Tester data={3} />
<Target />
```

`div(3)` is unmounted (it does not match to `{ other content }`), `div(2)` updates to `div(3)`, `div(1)` updates to `div(2)`, new `div(1)` is mounted.
Generally (for `n` *Tester* items), `(n - 1)` is unmounted, `[(n - 2), 0]` are updated to `[(n - 1), 1]`, `0` is mounted.

Since first element is remounted there can some side effects (like an animation) though according to the code first element is not expected to change in any way.


----
When this

```jsx
<Tester data={1} />
<Tester data={2} />
<Tester data={3} />
<Target />
```

changes to

```jsx
<Tester data={1} />
<Tester data={2} />
<Tester data={3} />
```

`div(3)` is mounted (`{ other content }` does not match to it), `div(3)` changes ti `div(2)`, `div(2)` changes to `div(1)`, `div(1)` is unmounted.
Generally (for `n` *Tester* items), `(n - 1)` is mounted, `[(n - 1), 1]` are updated to `[(n - 2), 0]`.

Since last element is remounted there can some side effects (like an animation) though according to the code first element is not expected to change in any way.

PR #1984, opened 2019-04-19
chore(react-chart): prevent scroll event on zooming

Krijovnick:


packages/dx-react-chart/src/types/plugins.zoom-pan.types.ts
```diff
@@ -35,3 +35,14 @@ export namespace ZoomAndPan {
     rect: Rect;
   }
 }
+
+type EventTriggerFn = (e) => void;
```
DmitryBogomolov: There is `EventHandlerFn` type (already exits).

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -51,37 +121,35 @@ class ZoomAndPanBase extends React.PureComponent<ZoomAndPanProps, ZoomAndPanStat
...
-    };
+  handleTouchMove(scales: ScalesCache, e: any) {
+    if (e.touches && e.touches.length === 2) {
+      const current = getDeltaForTouches(e.touches);
+      this.zoom(scales, current.delta - this.multiTouchDelta!, current.center);
+      this.multiTouchDelta = current.delta;
+    } else {
+      this.handleMouseMove(scales, { x: e.touches[0].clientX, y: e.touches[0].clientY });
+    }
+    e.preventDefault();
```
DmitryBogomolov: It usually does first in callback (unless called only under certain conditions).

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -173,34 +240,28 @@ class ZoomAndPanBase extends React.PureComponent<ZoomAndPanProps, ZoomAndPanStat
...
-            {({ scales }) =>
+            {({ scales, rootRef }) =>
             <DragDropProvider>
               <DropTarget
                 onOver={({ _, clientOffset }) => this.handleMouseMove(scales, clientOffset)}
                 onDrop={() => this.handleMouseUp(scales)}
               >
                 <DragSource payload={null}>
-                  <TemplatePlaceholder />
+                  <ZoomPanProvider
```
DmitryBogomolov: It does not need to render anything (its purpose is to access *mount* and *unmount* stages). So it should be moved out of *DragDropProvider* (hence be independent of provider's renders).

```jsx
<ZoomPanProvider ... />
<DragDropProvider ... />
```

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -18,9 +18,79 @@ import {
...
 
+const events = {
+  wheel: 'mouseWheel',
+  onDown: 'onDown',
+  touchmove: 'touchMove',
+  touchend: 'touchEnd',
+};
+
+class ZoomPanProvider extends React.PureComponent<ZoomPanProviderProps> {
+  static defaultProps: Partial<ZoomPanProviderProps> = {
```
DmitryBogomolov: It is not required (outer component always providers all props), isn't it?

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -18,9 +18,79 @@ import {
...
+    Object.keys(events).forEach((el) => {
+      rootRef.current!.removeEventListener(el, this[events[el]]);
+    });
+  }
+
+  componentWillUnmount() {
+    this.detachDocumentEvents();
+  }
+
+  mouseWheel = (e) => {
```
DmitryBogomolov: These handlers are all similar.

```jx
this.handlers = LIST.map(({ name, prop }) => ({
  name,
  handler: (e: any) => this.props[prop](e),
}));
```

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -18,9 +18,79 @@ import {
...
+    onTouchMove: (args) => {},
+    onTouchEnd: (args) => {},
+  };
+
+  componentDidMount() {
+    this.detachDocumentEvents();
+    this.attachDocumentEvents();
+  }
+
+  componentDidUpdate() {
```
DmitryBogomolov: Is it required?

Krijovnick: no, now it is not required

packages/dx-react-chart/src/plugins/zoom-pan.tsx
Krijovnick: React classes should render something, in this case it is `TemplatePlaceholder`. If `ZoomPanProvider` renders `null`, zooming wouldn't be work.

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -18,9 +18,55 @@ import {
...
+const events = {
+  wheel: 'onWheel',
+  mousedown: 'onDown',
+  touchstart: 'onDown',
+  touchmove: 'onTouchMove',
+  touchend: 'onTouchEnd',
+};
+
+class ZoomPanProvider extends React.PureComponent<ZoomPanProviderProps> {
+  ref: Element | undefined = undefined;
```
DmitryBogomolov: It should be declared as

```js
ref!: Element;
```

as we are sure that it will be available when required.

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -18,9 +18,55 @@ import {
...
+  componentDidMount() {
+    this.ref = this.props.rootRef.current!;
+    if (!this.ref) return;
```
DmitryBogomolov: How can it be `null`?

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -18,9 +18,55 @@ import {
...
+    this.detachDocumentEvents();
```
DmitryBogomolov: Why is it called?

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -18,9 +18,55 @@ import {
...
+
+  attachDocumentEvents() {
+    Object.keys(events).forEach((el) => {
+      this.ref!.addEventListener(el, this.handler(el), { passive: false });
+    });
+  }
+
+  detachDocumentEvents() {
+    Object.keys(events).forEach((el) => {
+      this.ref!.removeEventListener(el, this.handler(el));
```
DmitryBogomolov: Created handlers should be persisted in the instance - now `handler` creates a new instance on every call.

PR #2001, opened 2019-04-26
fix(react-grid): add transform style for IE11

churkin:


MaximKudriavtsev:
Should we add the same changes into this [file](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-scheduler-core/src/utils.ts#L198)?

packages/dx-react-grid/src/plugins/integrated-selection.tsx
```diff
@@ -36,6 +36,10 @@ class IntegratedSelectionBase extends React.PureComponent<IntegratedSelectionPro
         name="IntegratedSelection"
         dependencies={pluginDependencies}
       >
+        <Getter name="tableColumns" computed={({ selection, tableColumns }) =>  {
+          console.log('!!!!');
+          return tableColumns;
```
MaximKudriavtsev: This is an excess code

Issue #1971, opened 2019-04-14
Error when adding new series elements


alexkirsz:
- [x] I have searched this repository's [issues](https://github.com/devexpress/devextreme-reactive/issues) and believe that this is not a duplicate.

### I'm using ...

- [ ] React Grid
- [x] React Chart
- [ ] React Scheduler

### Current Behaviour

Adding new series after rendering the chart for the first time throws an error.

### Expected Behaviour

New series should be correctly added to the chart.

### Steps to Reproduce
https://codesandbox.io/s/vnqlwxlp0l

### Environment
See CodeSandbox

DmitryBogomolov:
Hello.

We need time to research the issue.
For now use the [following workaround](https://codesandbox.io/s/8yx9yo3900).

DmitryBogomolov:
[Sample](https://codesandbox.io/s/8zrj7mr2k9) for debugging.

DmitryBogomolov:
Closed via https://github.com/DevExpress/devextreme-reactive/pull/2002.

PR #2002, opened 2019-04-27
fix(react-core): prevent Getter losing on Plugin adding

DmitryBogomolov:
## What happens

`Plugin`, `Getter` and `Template` have internal objects (`this.plugin`) which ordered within `PluginHost` internal object (`this.plugins`) based on corresponding *positions*.
Position originates from component index within its parent children.

For this tree

```jsx
<PluginHost>
  <PluginA />
  <PluginC />
</PluginHost>
```

the list is the following

```js
[ A(0), C(1) ]
```

When tree changes to

```jsx
<PluginHost>
  <PluginA />
  <PluginB />
  <PluginC />
</PluginHost>
```

`PluginB` is created (internal object is created and added to list) and `PluginC` is updated (index changes).
During `PluginB` creation its internal object is created and has position `(1)`. But `PluginC` is not updated by that moment and its internal object position is `(1)` too.
Such position collision is resolved in such a way that `B(1)` replaces `C(1)`. The list is

```js
[ A(0), B(1) ]
```

----
Here is a [sample](https://codesandbox.io/s/py8k0qwy50) of React behavior.
Click the *Change* button and note that *C2* and *C3* are created *C4* and *C5* have outdated indexes.

## What is done

- `key` property is used to force recreation all items going after newly added
- `unregisterPlugin` fixed so that it does not accidentally remove last list item

Because of `key` `PluginC` is destroyed (and `C(1)` is remove from list) and created (and `C(2)` is added to to list) after `PluginB` is created. So the list is

```js
[ A(0), B(1), C(2) ]
```

Since `registerPlugin` allows the case when other list entry is removed (when there is position collision), `unregisterPlugin` should allow the case when no entry is removed (because it could already be removed in `registerPlugin`).

### DIFF
--- a/packages/dx-core/src/utils.ts
+++ b/packages/dx-core/src/utils.ts
@@ -23,8 +23,8 @@ export const insertPlugin = (array, newItem) => {
 
 /** @internal */
 export const removePlugin = (array, item) => {
-  const i = array.indexOf(item);
-  return i >= 0 ? [...array.slice(0, i), ...array.slice(i + 1)] : array;
+  const itemIndex = array.indexOf(item);
+  return itemIndex >= 0 ? [...array.slice(0, itemIndex), ...array.slice(itemIndex + 1)] : array;
 };
 
 /** @internal */
--- a/packages/dx-core/src/utils.ts
+++ b/packages/dx-core/src/utils.ts
@@ -1,7 +1,5 @@
-/** @internal */
 const DELAY = 200;
 
-/** @internal */
 const compare = (a, b) => {
   const aPosition = a.position();
   const bPosition = b.position();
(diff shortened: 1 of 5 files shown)


DmitryBogomolov:
Fixes https://github.com/DevExpress/devextreme-reactive/issues/1971.

packages/dx-core/src/utils.ts
```diff
@@ -23,6 +21,12 @@ export const insertPlugin = (array, newItem) => {
   return result;
 };
 
+/** @internal */
+export const removePlugin = (array, item) => {
+  const i = array.indexOf(item);
+  return i >= 0 ? [...array.slice(0, i), ...array.slice(i + 1)] : array;
```
MaximKudriavtsev: Can we change the `i` name to `itemIndex` or something more clearly?

DmitryBogomolov: Yes, we can.

PR #2006, opened 2019-04-30
fix(react-chart): fix spacing on axis labels in Edge

Krijovnick:
BREAKING CHANGE:

We have replaced the `dominantBaseline` option with the `dy` option for axis labels because Edge does not support the former. The `dy` option does not change the baseline position and simply shifts text up or down.


packages/dx-react-chart/src/templates/axis/label.tsx
```diff
@@ -4,12 +4,12 @@ import { Axis } from '../../types';
 export class Label extends React.PureComponent<Axis.LabelProps> {
   render() {
     const {
-      text, x, y, dominantBaseline, textAnchor, ...restProps
+      text, x, y, dy, textAnchor, ...restProps
     } = this.props;
 
     return (
```
DmitryBogomolov: Can't it be replace with this?

```jsx
const { text, ...restProps } = this.props;
return <text {...restProps}>{text}</text>;
```

packages/dx-react-chart/docs/reference/value-axis.md
```diff
@@ -70,7 +70,7 @@ Field | Type | Description
 text | string &#124; number | The label text.
 x | number | The x coordinate of the label's top left corner.
 y | number | The y coordinate of the label's top left corner.
-dominantBaseline | 'hanging' &#124; 'middle' &#124; 'baseline' | The label's baseline.
+dy | string | The label's offset from baseline in css units.
```
RomanTsukanov:
```suggestion
dy | string | The label's offset from the baseline in CSS units.
```

packages/dx-react-chart/docs/reference/argument-axis.md
```diff
@@ -69,7 +69,7 @@ Field | Type | Description
 text | string &#124; number | The label text.
 x | number | The x coordinate of the label's top left corner.
 y | number | The y coordinate of the label's top left corner.
-dominantBaseline | 'hanging' &#124; 'middle' &#124; 'baseline' | The label's baseline.
+dy | string | The label's offset from baseline in css units.
```
RomanTsukanov:
```suggestion
dy | string | The label's offset from the baseline in CSS units.
```

packages/dx-react-chart/src/types/templates.axis.types.ts
```diff
@@ -9,8 +9,8 @@ export namespace Axis {
     x: number;
     /** The y coordinate of the label’s top left corner */
     y: number;
-    /** The label’s baseline */
-    dominantBaseline: string;
+    /** The label's offset from baseline in css units */
```
RomanTsukanov:
```suggestion
    /** The label's offset from the baseline in CSS units */
```

PR #2020, opened 2019-05-13
chore(react-chart): fix cutting of the last grid

Krijovnick:


packages/dx-react-chart/src/types/plugins.pane-layout.types.ts
```diff
@@ -0,0 +1,4 @@
+/** @internal */
+export interface PaneLayoutProps {
```
DmitryBogomolov: Isn't is unused?

packages/dx-react-chart/src/plugins/pane-layout.tsx
```diff
@@ -36,6 +39,8 @@ export class PaneLayout extends React.PureComponent {
             <TemplateConnector>
               {({ layouts }, { changeBBox }) => {
                 const { width, height } = layouts.pane;
+                const clipPathId = `clip_path_${numDefs}`;
```
DmitryBogomolov: Unique clip path is bound to a component instance not to a particular template render occurrence.

packages/dx-react-chart/src/types/plugins.pane-layout.types.ts
Krijovnick: yes, i'll remove it

packages/dx-react-chart/src/plugins/pane-layout.tsx
DmitryBogomolov: I meant - *should be* bound.

PR #2022, opened 2019-05-14
chore(react-chart): fix default tick count calculation

DmitryBogomolov:


packages/dx-chart-core/src/plugins/axis/computeds.test.ts
```diff
@@ -130,6 +130,24 @@ describe('axisCoordinates', () => {
...
+    it('should generate ticks when pane size is zero, horizontal', () => {
+      (isHorizontal as jest.Mock).mockReturnValue(true);
+      const coordinates = axisCoordinates({
+        scale, tickSize, indentFromAxis, scaleName: 'test-name', position: 'top',
+        paneSize: [0, 0],
+      } as any);
+      expect(scale.ticks).toBeCalledWith(10);
+    });
+
+    it('should generate ticks when pane size is zero, vertical', () => {
```
Krijovnick: One of the test is excess

DmitryBogomolov: All tests of **linear** module are paired for horizontal and vertical cases, except for tests for formatting.
Tick count calculation depends on orientation in the sense that `isHor` flag is used to determine which part of the pane size object is used.

Krijovnick: On each new part of code we write one test. There are no reference about orientation in the changes. And tests with dependence of `paneSize` and orientation has already. Thus we should only check that the function returns the ticks if `paneSize` has zero value. 
By the way, constant `coordinates` is also reduntal.

DmitryBogomolov: Though there is no orientation reference in the changes, tick calculation depends on the orientation.

```js
const tickCount = getTickCount(scale.range(), paneSize[1 - Number(isHor)]);
```
Should **should pass correct domain to scale, vertical** test be removed?

Krijovnick: I think it is not matter vertical or horizontal test will be removed. In both cases we will be in the same part of code.

DmitryBogomolov: I meant that there are two **should pass correct domain to scale** tests. And it one of tests in this **PR** is to be removed then one of **pass correct domain** tests should be removed as well.
Shouldn't it?

Krijovnick: yes

DmitryBogomolov: Let it be so then.

PR #2029, opened 2019-05-15
chore: update deps

MaximKudriavtsev:


packages/dx-react-chart/src/plugins/spline-series.test.tsx
```diff
@@ -48,24 +48,28 @@ describe('Spline series', () => {
...
-      pointComponent: undefined,
-      index: 1,
-      coordinates: coords,
-      color: 'color',
-      getAnimatedStyle: undefined,
-      scales: { xScale: 'arg-scale', yScale: 'val-scale' },
-    });
   });
+
+  // it('should render path', () => {
```
ushkal: Does SplineSeries chart work?

MaximKudriavtsev: Yes, it does. But the test doesn't work.

PR #2037, opened 2019-05-21
feat(react-grid): stick group row on horizontal scroll

ushkal:


packages/dx-react-grid-bootstrap3/src/templates/table-group-cell/container.jsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState, useContext } from 'react';
+import * as PropTypes from 'prop-types';
+import { getStickyPosition } from '../../utils/css-fallback-properties';
+import { ThemeColors } from '../layout';
+
+export const Container = ({ children, style, ...restProps }) => {
+  const [position] = useState(getStickyPosition());
```
DmitryBogomolov: With `setState` result of `useState` omitted, `useState` is the same as just assignment, isn't it?

```jsx
const [y] = useState(x);
// or
const y = x;
```

Shouldn't it be something like that?
```jsx
const [position, setPosition] = useState(getStickyPosition());
useEffect(() => {
  setPosition(getStickyPosition());
}, []);
```

packages/dx-react-grid-bootstrap3/src/templates/table-group-cell/cell.jsx
```diff
@@ -17,19 +19,21 @@ export const Cell = ({
...
-      column={column}
-      row={row}
-    >
-      {children}
-    </Content>
+    <Container style={{ left }}>
+      <Icon
+        expanded={expanded}
+        onToggle={onToggle}
+        style={{
```
DmitryBogomolov: Property value is new on every `render` thought it actually never changes. Extract a constant.

packages/dx-react-grid-bootstrap3/src/templates/table-group-cell/container.jsx
ushkal: > With setState result of useState omitted, useState is the same as just assignment, isn't it?

Yes, this is intended - we need to initialize this variable only once on component render

DmitryBogomolov: Do you mean that `getStickyPosition` must be called only once in a component lifetime (like in `constructor`)?

ushkal: Yes, exactly - like in table fixed cell component

packages/dx-react-grid/src/plugins/table-group-row.tsx
```diff
@@ -106,12 +116,14 @@ class TableGroupRowBase extends React.PureComponent<TableGroupRowProps> {
...
                           contentComponent={Content}
                           iconComponent={Icon}
+                          containerComponent={Container}
                           row={params.tableRow.row}
                           column={params.tableColumn.column!}
                           expanded={expandedGroups.indexOf(params.tableRow.row.compoundKey) !== -1}
                           onToggle={() => toggleGroupExpanded(
                             { groupKey: params.tableRow.row.compoundKey },
                           )}
+                          left={contentLeft}
```
MaximKudriavtsev: Can we rename this property and others similar? If we are going to create an RTL support for Grid, this property should be renamed.

We can use `indent` name instead.

packages/dx-react-grid-material-ui/src/templates/table-group-cell/indent-cell.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+const IndentCellBase = ({
+  left,
+  tableRow,
+  tableColumn,
+  row, column,
+  style, className, classes,
+  ...restProps
+}) => (
+  <td
```
MaximKudriavtsev: Why we use `td` instead of mui TableCell here?

packages/dx-react-grid-material-ui/src/templates/table-group-cell/container.jsx
```diff
@@ -0,0 +1,39 @@
...
+    display: 'inline-block',
+  },
+});
+
+const ContainerBase = ({
+  children, style, classes, className, ...restProps
+}) => (
+  <div
+    className={classNames(classes.wrapper, className)}
+    style={style}
```
MaximKudriavtsev: Should we use the following syntax instead of current?

```jsx
const ContainerBase = ({
  children, left, classes, className, ...restProps
}) => (
  <div
    className={classNames(classes.wrapper, className)}
    style={{ left, ...style }}
    ...
```

packages/dx-react-grid-bootstrap3/src/templates/table-group-cell/container.jsx
DmitryBogomolov: It is rather strange - `getStickyPosition` will be called on each component update (without purpose) though the result of the first call will always be used.
https://codesandbox.io/s/billowing-glitter-ij0my

As for `useEffect` - it seems that it can be omitted. And `componentDidMount` in `FixedCell` can be omitted as well.

----
But actually - both should be kept.
`getStickyPosition` is not pure (it accesses DOM) - hence it should not be called neither from `render` nor from `constructor`.
Sticky position usage seems to be like this - state field is initialized with `undefined` (or something like this, but not with `getStickyPosition()`) and on *mount* stage field is updated with `getStickyPosition()`. Doesn't it?

ushkal: We better move it to a context and calculate it once per page load - now a sticky position gets calculated once per row at best

packages/dx-react-grid-bootstrap4/src/templates/table-group-cell/cell.jsx
```diff
@@ -1,12 +1,15 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { argumentsShallowEqual } from '@devexpress/dx-core';
 
-export const Cell = ({
+export const Cell = React.memo(({
```
DmitryBogomolov: `children` property will be different (most likely) on each new render, won't it?

packages/dx-react-grid/docs/reference/table-group-row.md
```diff
@@ -39,6 +39,7 @@ contentComponent | ComponentType&lt;[TableGroupRow.ContentProps](#tablegrouprowc
 iconComponent | ComponentType&lt;[TableGroupRow.IconProps](#tablegrouprowiconprops)&gt; | | A component that renders a group expand icon.
 indentCellComponent? | ComponentType&lt;[TableGroupRow.IndentCellProps](#tablegrouprowindentcellprops)&gt; | null | A component that renders a group indent cell.
 indentColumnWidth | number | | The group indent column's width.
+contentCellPadding | string | | The group cell's left padding value.
```
MaximKudriavtsev: Should we define a `default` value for all themes, as we did [here](https://devexpress.github.io/devextreme-reactive/react/grid/docs/reference/virtual-table/)? And `indentColumnWidth` too.

packages/dx-react-grid-bootstrap4/src/templates/table-group-cell/cell.jsx
ushkal: That's right, we shouldn't use memo here

PR #2038, opened 2019-05-22
chore: mechanism to decrease code duplication in demos

DmitryBogomolov:


packages/dx-demo-shell/generate-demos.js
```diff
@@ -121,18 +136,31 @@ const overrideFileIfChanged = (filename, data) => {
...
 };
+const findSplitter = (content, index) => {
+  const k = content.indexOf(' from ', index + 1);
+  if (k >= 0) {
+    return findSplitter(content, k);
+  }
+  return index >= 0 ? content.indexOf('\n', index + 1) : -1;
+};
+const splitPartialContent = (content) => {
+  const index = findSplitter(content, -1);
```
ushkal: Maybe it'd be more clear if we used some splitter tag in partial templates

DmitryBogomolov: Yes, it would. But we will have to manually put slitter tag in all files.
We have to choose.

Let everybody vote here.

packages/dx-demo-shell/generate-demos.js
```diff
@@ -121,18 +136,31 @@ const overrideFileIfChanged = (filename, data) => {
 };
 const createFromTemplate = (sourceFilename, outputFilename, data) => {
   const source = fs.readFileSync(sourceFilename, 'utf-8');
-  mustache.tags = ['<%', '%>'];
   const output = mustache.render(source, data);
   overrideFileIfChanged(outputFilename, output);
   cancelFileRemoving(outputFilename);
 };
+const findSplitter = (content, index) => {
+  const k = content.indexOf(' from ', index + 1);
```
Krijovnick: What about using `lastIndexOf` instead of `indexOf`?

DmitryBogomolov: It is slightly more dangerous.
Searching from the beginning of a file chance that `__from__` is part of `import` statement is quite high.
Searching from the ending of a file `__from__` may be part of some title value.
```jsx
<SomePlugin title="Some data from some source" />
```

However, if there is no `import` section and there is some title with *from* word current search would not work as well.

----
Current search needs improvement.
And it is `+1` for slitter tag option.

packages/dx-demo-shell/generate-demos.js
Krijovnick: +1 to use slitter tag

DmitryBogomolov: Now each block is expected to be wrapped with 
```jsx
// BLOCK:<block-name>
...
// BLOCK:<block-name>
```
and then referenced as `<%&block-name%>`.

PR #2039, opened 2019-05-22
chore(react-chart): refactor zoom and pan plugin

Krijovnick:


packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -45,13 +43,13 @@ class ZoomPanProvider extends React.PureComponent<ZoomPanProviderProps> {
 
   attachEvents() {
     Object.keys(this.handlers).forEach((el) => {
-      this.ref.addEventListener(el, this.handlers[el], { passive: false });
+      window.addEventListener(el, this.handlers[el], { passive: false });
```
DmitryBogomolov: Event handlers should not be attached to `window` during all the time of widget existence.
For `onWheel` it is excessive - we need **wheel** events only from widget container.
For `onMouseMove` it is potentially dangerous - because of unconditional `preventDefault`.

**wheel** and **start** handlers should be attached to widget container.
**move** and **end** should be attached to `window` but only during the time the corresponding gesture is being processed - attached on **start** and detached when finished (on **end** or due to other reason).

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -95,49 +93,46 @@ class ZoomAndPanBase extends React.PureComponent<ZoomAndPanProps, ZoomAndPanStat
...
-      // "handleMouseMove" and "handleMouseUp".
+      // "handleMove" and "handleEnd".
       // TODO: Provide rectangle mode canceling.
     if (isKeyPressed(e, zoomRegionKey)) {
       this.rectOrigin = [e.pageX - this.offset[0], e.pageY - this.offset[1]];
     }
     if (e.touches && e.touches.length === 2) {
       this.multiTouchDelta = getDeltaForTouches(e.touches).delta;
     }
+    this.lastCoordinates = [e.clientX - this.offset[0], e.clientY - this.offset[1]];
```
DmitryBogomolov: Somewhere `pageX` and `pageY` are used (here in this plugin and in `EventTracker`) and somewhere - `clientX` and `clientY`.
Can we bring it to single variant? Looks like it should be `pageX` and `pageY`.

Krijovnick: ok, let's make `pageX` and `pageY`

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -15,48 +12,66 @@ import {
...
-  detachEvents() {
-    Object.keys(this.handlers).forEach((el) => {
-      this.ref.removeEventListener(el, this.handlers[el]);
+  detachEvents(node, handlers) {
+    Object.keys(handlers).forEach((el) => {
+      node.removeEventListener(el, handlers[el]);
     });
   }
 
   componentWillUnmount() {
```
DmitryBogomolov: There may be events attached to `window` - they should be detached as well.
(It may happen when mouse is pressed and widget is destroyed before mouse is released)

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -97,47 +112,39 @@ class ZoomAndPanBase extends React.PureComponent<ZoomAndPanProps, ZoomAndPanStat
...
-  handleTouchMove(scales: ScalesCache, e: any) {
+  handleMove(scales: ScalesCache, e: any) {
     e.preventDefault();
     if (e.touches && e.touches.length === 2) {
       const current = getDeltaForTouches(e.touches);
       this.zoom(scales, current.delta - this.multiTouchDelta!, current.center);
       this.multiTouchDelta = current.delta;
     } else {
-      this.handleMouseMove(scales, { x: e.touches[0].clientX, y: e.touches[0].clientY });
+      this.scroll(scales, e.touches ? e.touches[0] : e);
```
DmitryBogomolov: The `e.touches ? e.touches[0] : e` code is the same as that in event coordinates calculation. Can't it be replaced with this?
```jsx
this.scroll(scales, getEventCoords(e));
```

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
```diff
@@ -153,3 +153,9 @@ export const isKeyPressed = (event: MouseEvent, key: string) => event[`${key}Key
 export const getWheelDelta = ({ wheelDelta, deltaY }: { wheelDelta?: number, deltaY?: number }) => {
   return wheelDelta !== undefined ? wheelDelta : deltaY! * -30; // deltaY for FF
 };
+
+/** @internal */
+export const getCoordsWithOffset = (e: any, offset: NumberArray): NumberArray => {
+  const { pageX, pageY } = e.touches ? e.touches[0] : e;
```
DmitryBogomolov: Let's extract a function.
```jsx
const getEventCoords = (e: any): NumberArray => {
  const { pageX, pageY } = e.touches ? e.touches[0] : e;
  return [pageX, pageY];
};
```
That's because similar code is used in `EventTracker`.

And `getCoordsWithOffset` is also used in `EventTracker`.

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -97,47 +112,39 @@ class ZoomAndPanBase extends React.PureComponent<ZoomAndPanProps, ZoomAndPanStat
...
       // Rectangle mode should be canceled if "zoomRegionKey" is released during mouse movevent or
       // not pressed when mouse is up. To do it access to "event" object is required in
-      // "handleMouseMove" and "handleMouseUp".
+      // "handleMove" and "handleEnd".
       // TODO: Provide rectangle mode canceling.
     if (isKeyPressed(e, zoomRegionKey)) {
-      this.rectOrigin = [e.pageX - this.offset[0], e.pageY - this.offset[1]];
+      this.rectOrigin = coords;
     }
     if (e.touches && e.touches.length === 2) {
```
DmitryBogomolov: Let's move it to *dx-chart-core* package - something like `isMultiTouch`.

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -192,10 +199,10 @@ class ZoomAndPanBase extends React.PureComponent<ZoomAndPanProps, ZoomAndPanStat
     });
   }
 
-  handleScroll(scales: ScalesCache, e: any) {
+  handleZoom(scales: ScalesCache, e: any) {
     e.preventDefault();
     const offset = getOffset(e.currentTarget);
-    const center: NumberArray = [e.pageX - offset[0], e.pageY - offset[1]];
+    const center: NumberArray = getCoordsWithOffset(e, offset);
```
DmitryBogomolov: There is no need to specify type - it is inferred from function declaration.

packages/dx-react-chart/src/plugins/zoom-pan.tsx
Krijovnick: No, there is no `getBoundingClientRect` in event.

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -15,48 +12,80 @@ import {
...
       };
     }, {});
-    this.attachEvents();
+    this.attachEvents(this.ref, this.handlers);
   }
 
-  attachEvents() {
-    Object.keys(this.handlers).forEach((el) => {
-      this.ref.addEventListener(el, this.handlers[el], { passive: false });
+  attachEvents(node, handlers) {
```
DmitryBogomolov: This is not an instance method (should be moved out of the class).

packages/dx-react-chart/src/plugins/zoom-pan.tsx
```diff
@@ -15,48 +12,80 @@ import {
...
+    Object.keys(handlers).forEach((el) => {
+      node.addEventListener(el, handlers[el], { passive: false });
     });
   }
 
-  detachEvents() {
-    Object.keys(this.handlers).forEach((el) => {
-      this.ref.removeEventListener(el, this.handlers[el]);
+  detachEvents(node, handlers) {
```
DmitryBogomolov: This is not an instance method.

PR #2040, opened 2019-05-22
fix(react-scheduler): update views by passing new props

### LINKED ISSUE #2026: Override WeekView to re-render when its props change

### I'm using ...

- [ ] React Grid
- [ ] React Chart
- [x] React Scheduler

### Description

Hi, it seems like the current implementation of WeekView in React Scheduler has the following constraint: 

- **props such as startDayHour, endDayHour and excludedDays** etc. only gets rendered once, and **doesn't re-render after the first initialisation**

Am I correct? If so, how do I override such behaviour, as I would like to allow users to dynamically update startDayHour and endDayHour that fits their needs (e.g some people might not have plans from 12am-6am, but some might). 

My current implementation is as follows:
- a parent component that holds the states that can be manipulated by users (e.g. startDayHour); 
- a child component that includes Week View, such as:
`<WeekView
  startDayHour={this.props.startDayHour}
/>`
, where the props are managed by parent component
- when the user update the startDayHour via an input field, parent's state is updated 
- WeekView startDayHour is updated accordingly, but doesn't get re-rendered despite the value is changed 

This is probably due to the **current implementation of WeekView as a Pure Component, how can I get around this**?

Please correct me if I'm wrong. Thanks in advance!

### Environment

- devextreme-reactive: xxx
- react: xxx
- browser: xxx
- bootstrap: none
- react-bootstrap: none
- material-ui: none

MaximKudriavtsev:
Fixes #2026


packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/time-table/layout.jsx
```diff
@@ -12,49 +12,84 @@ const styles = {
...
+
+  componentDidMount() {
+    const { setCellElements } = this.props;
+
+    const cellElements = this.table.querySelectorAll('td');
+    setCellElements(cellElements);
+  }
+
+  componentDidUpdate() {
+    const { setCellElements } = this.props;
```
ushkal: A function can be extracted

packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/time-table/layout.test.jsx
```diff
@@ -37,29 +45,41 @@ describe('Horizontal view TimeTable', () => {
...
+          {...defaultProps}
+        />
+      ));
+
+      expect(defaultProps.setCellElements)
+        .toBeCalledTimes(1);
+      expect(defaultProps.setCellElements)
+        .toHaveBeenCalledWith(expect.arrayContaining([]));
+
+      tree.setProps({ a: 1 });
```
ushkal: mockClear should be called between assertions

packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/time-table/layout.jsx
```diff
@@ -12,49 +12,84 @@ const styles = {
...
+      classes,
+      tableRef,
+      className,
+      cellsData,
+      formatDate,
+      setCellElements,
+      ...restProps
+    } = this.props;
+    return (
+      <RootRef rootRef={this.saveReference}>
```
ushkal: Can we extract `RootRef -> Table -> Body` to a container component to be used in both horizontal and vertical layouts?

packages/dx-react-scheduler/src/plugins/day-view.tsx
```diff
@@ -70,79 +71,79 @@ class DayViewBase extends React.PureComponent<VerticalViewProps, ViewState> {
...
   layoutHeaderElementComputed = () => this.layoutHeader;
 
-  layoutHeaderElement: ComputedFn = (getters) => {
-    const { name: viewName } = this.props;
+  layoutHeaderElement = viewName => (getters) => {
     return computed(
       getters, viewName!, this.layoutHeaderElementComputed, getters.layoutHeaderElement,
     );
   }
+  memoLayoutHeaderElement = memoize(this.layoutHeaderElement);
```
ushkal: It's better to rename memo -> memoized to avoid confusion with React.Memo

MaximKudriavtsev: I used the `memo` word to reduce length.

packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/time-table/layout.jsx
MaximKudriavtsev: No, because our Scheduler consists of 2/3 (horizontal/vertical) tables.

ushkal: This container component is supposed to be rendered in each view as it is now. I suggest to extract this refs logic:
`RootRef -> Table -> Body -> Row`
 to
`LayoutContainer -> Row`

packages/dx-react-scheduler/src/plugins/day-view.tsx
```diff
@@ -179,23 +207,31 @@ class DayViewBase extends React.PureComponent<VerticalViewProps, ViewState> {
...
+        />
         <Getter name="startViewDate" computed={this.startViewDateComputed} />
         <Getter name="endViewDate" computed={this.endViewDateComputed} />
-        <Getter name="timeTableElement" computed={this.timeTableElement} />
-        <Getter name="layoutElement" computed={this.layoutElement} />
-        <Getter name="layoutHeaderElement" computed={this.layoutHeaderElement} />
+
+        <Getter name="timeTableElement" computed={this.memoTimeTableElement(viewName)} />
+        <Getter name="layoutElement" computed={this.memoLayoutElement(viewName)} />
+        <Getter name="layoutHeaderElement" computed={this.memoLayoutHeaderElement(viewName)} />
```
DmitryBogomolov: Memoization is expected to be transparent - calling side should not distinguish if it calls memoized function or not.

Not
```jsx
<Getter name="something" computed={this.memoGetSomethingCalculator(...)} />
```

But
```jsx
<Getter name="something" computed={this.getSomethingCalculator(...)} />
```

----
Consequently - there is no need to keep original function together with memoized function in component instance.

Not
```jsx
getSomethingCalculator = (...) => { 
  ...
};
...
memoGetSomethingCalculator = memoize(this.getSomethingCalculator);
```

But
```jsx
getSomethingCalculator = memoize((...) => {
  ...
});
```

packages/dx-react-scheduler-material-ui/src/templates/views/common/time-table/layout-container.jsx
```diff
@@ -0,0 +1,51 @@
...
+import * as PropTypes from 'prop-types';
+import RootRef from '@material-ui/core/RootRef';
+
+export class TimeTableContainer extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.table = null;
+
+    this.setCells = this.setCells.bind(this);
```
DmitryBogomolov: There is no need in `bind`.

packages/dx-react-scheduler-material-ui/src/templates/views/common/time-table/layout-container.jsx
```diff
@@ -0,0 +1,51 @@
...
+    this.saveReference = this.saveReference.bind(this);
```
DmitryBogomolov: Why not [field declaration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes#Field_declarations)?

MaximKudriavtsev: What do you mean?

DmitryBogomolov:
```jsx
class TimeTableContainer ... {
  ...
  saveReference = (ref) => {
    const { tableRef } = this.props;
    this.table = ref;
    tableRef(ref);
  };
  ...
}
```

ushkal: As I know, it works in TS only

DmitryBogomolov: Yes, I forgot that it is another package.
I withdraw the question.

packages/dx-react-scheduler-material-ui/src/templates/views/common/time-table/layout-container.jsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import RootRef from '@material-ui/core/RootRef';
+
+export class TimeTableContainer extends React.PureComponent {
```
DmitryBogomolov: Since component has `children` property (which most likely will be different on each render) it should extend `React.Component` (not `React.PureComponent`).

packages/dx-react-scheduler/src/plugins/day-view.tsx
```diff
@@ -161,9 +154,36 @@ class DayViewBase extends React.PureComponent<VerticalViewProps, ViewState> {
...
+        cellDuration,
+        currentDate,
+        viewCellsData,
+        cellElements,
+      },
+    );
+
+    this.setState({ rects });
+  }
+  memoCalculateRects =  memoize(this.calculateRects);
```
DmitryBogomolov: Memoization transparency.

packages/dx-react-scheduler/src/plugins/day-view.tsx
```diff
@@ -71,79 +72,71 @@ class DayViewBase extends React.PureComponent<VerticalViewProps, ViewState> {
...
       getters, viewName!, this.timeTableElementComputed, getters.timeTableElement,
     );
-  }
+  });
 
-  viewCellsDataComputed: ComputedFn = (getters) => {
-    const { name: viewName } = this.props;
-    return computed(
-      getters, viewName!, this.viewCellsDataBaseComputed, getters.viewCellsData,
+  viewCellsDataBaseComputed = (startDayHour, endDayHour, cellDuration) => ({
```
DmitryBogomolov: This function should be a method not an instance function (because it is called with proper context).

packages/dx-react-scheduler/src/plugins/month-view.tsx
```diff
@@ -104,51 +95,76 @@ class MonthViewBase extends React.PureComponent<MonthViewProps, ViewState> {
...
+  viewCellsDataComputed: ComputedFn = (getters) => {
+    return computed(
+      getters,
+      getters.viewName!,
+      this.viewCellsDataBaseComputed(getters.firstDayOfWeek, getters.intervalCount),
+      getters.viewCellsData,
+    );
+  }
+
   setTimeTableRef = (timeTableRef) => {
```
DmitryBogomolov: With `setState` removed it can be replaced with `React.ref`.

PR #2042, opened 2019-05-23
fix(react-grid): Fixed column cell of selected row is not highlighted

### LINKED ISSUE #1948: DataGrid - Fixed column cell of selected row is not highlighted
### Steps to Reproduce:

1. Take a simple Grid
2. Add TableSelection
3. Add TableFixedColumns
4. Add columns to TableFixedColumns

Example here: https://codesandbox.io/s/l51v63mj5m

### Results You Received:
All cells from selected row are highlighted, apart from the cell from the fixed column

### Results You Expected:
All cells from selected row to be highlighted

churkin:
close #1948

### DIFF
--- a/packages/dx-react-grid-bootstrap4/src/templates/table-select-row.jsx
+++ b/packages/dx-react-grid-bootstrap4/src/templates/table-select-row.jsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
+import classNames from 'classnames';
 
 export const TableSelectRow = ({
   selected,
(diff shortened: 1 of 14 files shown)


packages/dx-react-grid-bootstrap4/src/templates/table-select-row.jsx
```diff
@@ -1,16 +1,20 @@
...
   children,
   style,
   onToggle,
   selectByRowClick,
+  className,
 }) => (
   <tr
     style={style}
-    className={selected ? 'table-active' : ''}
+    className={classNames({
```
ushkal: We should test that custom className is applied in such places

PR #2043, opened 2019-05-23
feat(react-scheduler): add a capability to set Scheduler's height

MaximKudriavtsev:


packages/dx-react-scheduler-material-ui/src/templates/layouts/horizontal-view-layout.jsx
```diff
@@ -30,15 +29,18 @@ export class HorizontalViewLayoutBase extends React.PureComponent {
       classes,
       layoutRef,
       layoutHeaderRef,
+      height,
     } = this.props;
 
+    const containerStyle = height ? { height: `${height}px` } : undefined;
```
ushkal: Can we provide a way to set auto height?

MaximKudriavtsev: What do you mean by `auto`?

ushkal: I mean 100% of a container's height. I know it can be done using AutoSizer but if it's not hard we should try to implement it

MaximKudriavtsev: Stretch to parent element?

MaximKudriavtsev: Okay, we should try

packages/dx-react-scheduler/api/dx-react-scheduler.api.md
```diff
@@ -438,6 +438,7 @@ export namespace MonthView {
   // (undocumented)
   export interface LayoutProps {
     dayScaleComponent: React.ComponentType<MonthView.DayScaleLayoutProps>;
+    height: number;
```
ushkal: Should be `number | 'auto'` and in docs too

PR #2057, opened 2019-05-30
chore(react-chart): draw rect for zoom depends on interaction options

Krijovnick:


packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
```diff
@@ -171,3 +171,19 @@ export const detachEvents = (node: any, handlers: EventHandlers) => {
     node.removeEventListener(el, handlers[el]);
   });
 };
+
+/** @internal */
+export const getRect = (
+  interactionWithArguments, interactionWithValues, prevCoords, nextCoords, rootRef, offset,
+) => {
+  const bbox = rootRef.current.getBoundingClientRect();
```
DmitryBogomolov: Can't we take `(0, 0)` as origin and `layouts.pane` as size for full rectangle?

The other thing (if `getBoundingClientRect` is still required) - `this.offset` is cached and `getBoundingClientRect` is not while they actually have same basis. Either both should be cached or both should not cached.

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
```diff
@@ -171,3 +171,19 @@ export const detachEvents = (node: any, handlers: EventHandlers) => {
...
+  const isZoomArgument = interactionWithArguments === 'both' || interactionWithArguments === 'zoom';
```
DmitryBogomolov: We already have similar code (in `boundsForScale`).

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
```diff
@@ -171,3 +171,19 @@ export const detachEvents = (node: any, handlers: EventHandlers) => {
     node.removeEventListener(el, handlers[el]);
   });
 };
+
+/** @internal */
+export const getRect = (
```
DmitryBogomolov: Variable names - `prevCoords` and `nextCoords`. They are not previous and next - they are original point and current point.

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
Krijovnick: `layout.pane` provide incomplete information, only width and height. We need also `left` and `top` fields

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
```diff
@@ -171,3 +173,25 @@ export const detachEvents = (node: any, handlers: EventHandlers) => {
...
+  interactionWithArguments: Interaction,
+  interactionWithValues: Interaction,
+  initial: Location,
+  current: Location,
+  bbox: BBox,
+) => {
+  const isZoomArgument = checkInteraction(interactionWithArguments, 'none', 'pan');
```
ushkal: Seems that 'none', 'pan' can be moved inside the checkInteraction function

Krijovnick: https://github.com/DevExpress/devextreme-reactive/pull/2057/files#diff-ed254ffc4e96edd9af97d74765ab4c8cR86 also use `checkInteraction` function.

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
DmitryBogomolov: Left and top are both `0` - there are already similar assumptions in **Axis** grids.

packages/dx-chart-core/src/types/plugins.zoom-pan.types.ts
```diff
@@ -10,3 +10,8 @@ export type ViewportOptions = {
 };
 
 export type OnViewportChangeFn = (viewport: ViewportOptions) => void;
+
+/** @internal */
+export type Pane = { width: number, height: number };
```
DmitryBogomolov: This type already exists - it is type of `layout` Getter value field - `Size`.

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
```diff
@@ -174,16 +176,22 @@ export const detachEvents = (node: any, handlers: EventHandlers) => {
...
+  const x = isZoomArgument ? Math.min(initial[0], current[0]) : 0;
+  const width = isZoomArgument ? Math.abs(initial[0] - current[0]) : pane.width;
+  const y = isZoomValue ? Math.min(initial[1], current[1]) : 0;
+  const height = isZoomValue ? Math.abs(initial[1] - current[1]) : pane.height;
   return {
     x, y, width, height,
   };
 };
+
+const checkInteraction = (interaction, value1, value2) =>
```
DmitryBogomolov: There should be type definitions.
And it can be simplified - `(interaction, type) => interaction === 'both' || interaction === type`.

In `boundsForScale` interaction type is valid if it is `'both'` or matches `type`.
In `getRect` interaction type is valid if it is `'both'` or `'zoom'`.

Krijovnick: `(interaction, type) => interaction === 'both' || interaction === type` and `(interaction, type) => interaction !== 'both' && interaction !== type` - there is no critical differences and there is the same difficulty. It is really need to be changed?

DmitryBogomolov: The point is to reduce number of arguments - from `type1` and `type2` to `type` (because `'both'` is always used). The meaning is - *Does specified interaction type match expected interaction type?*.
As for `(=== || ===)` or `(!== && !==)` - it does not matter.

packages/dx-chart-core/src/plugins/zoom-pan/computeds.ts
```diff
@@ -171,3 +174,25 @@ export const detachEvents = (node: any, handlers: EventHandlers) => {
...
+  const x = isZoomArgument ? Math.min(initial[0], current[0]) : 0;
+  const width = isZoomArgument ? Math.abs(initial[0] - current[0]) : pane.width;
+  const y = isZoomValue ? Math.min(initial[1], current[1]) : 0;
+  const height = isZoomValue ? Math.abs(initial[1] - current[1]) : pane.height;
+  return {
+    x, y, width, height,
+  };
+};
+
+const checkInteraction = (interaction: Interaction, type: string) =>
```
DmitryBogomolov: `type: Interaction`

PR #2058, opened 2019-05-30
chore(react-chart): title and legend for demo

DmitryBogomolov:
Adds "real" data, title and legend for [this demo](https://devexpress.github.io/devextreme-reactive/react/chart/demos/more_features/zoom-and-pan/).


packages/dx-react-chart-demos/src/demo-sources/zoom-pan/scatter.partial.jsxt
```diff
@@ -0,0 +1,116 @@
...
+      <table>
+        <tr>
+          <td><TooltipCell>Life Expectancy</TooltipCell></td>
+          <td><TooltipCell>{arg}</TooltipCell></td>
+        </tr>
+        <tr>
+          <td><TooltipCell>Birth Rate</TooltipCell></td>
+          <td><TooltipCell>{val}</TooltipCell></td>
+        </tr>
+      </table>
```
Krijovnick: Maybe make tooltip design like in the [devextreme charts demo](https://js.devexpress.com/Demos/WidgetsGallery/Demo/Charts/Overview/jQuery/Light/): with bold text in the header and add opacity to the splitter, `Life Expectancy` and `Birth Rate` phrases?

packages/dx-react-chart-demos/src/demo-sources/zoom-pan/bootstrap4/scatter.partial.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+// BLOCK:body
+const TooltipCell = props => (
+  <span {...props} />
+);
+const TooltipSplitter = ({ className, ...props }) => (
+  <hr className={classnames('mt-1 mb-1', className)} {...props} />
+);
+const TooltipHeader = props => (
+  <h6 {...props} /> // eslint-disable-line jsx-a11y/heading-has-content
```
Krijovnick: Is there comment required?

DmitryBogomolov: The suppressed error is the following:
> Headings must have content and the content must be accessible by a screen reader

MaximKudriavtsev:
image
image

The values are mixed up

packages/dx-react-chart-demos/src/demo-sources/zoom-pan/scatter.partial.jsxt
DmitryBogomolov: Here is how it looks now.
pr2058_bs4
pr2058_mui

Headers visually are not so bold as those in *devextreme demo* but they have `font-weight: 500` like in *devextreme demo*. With `600` they would look rather ugly.

----
The amount of non-similarity between *bs4* and *mui* markup is TERRIBLE though.

MaximKudriavtsev:
BTW, can we write the scale titles? This point will be helpful to understand these data set.

DmitryBogomolov:
No, there is no builtin mechanism for axis title.
And adding it via overriding **Axis.rootComponent** does not seem an easy option by now.

PR #2065, opened 2019-06-05
refactor(react-grid): extract viewport getter for virtual table

ushkal:


packages/dx-grid-core/src/utils/virtual-table.ts
```diff
@@ -80,7 +79,7 @@ export const getColumnBoundaries: PureComputed<
   )
 );
 export const getRowsVisibleBoundary: PureComputed<
-[TableRow[], number, number, GetColumnWidthFn, number, number, boolean?], RowsVisibleBoundary
+[TableRow[], number, number, GetColumnWidthFn, number, number, boolean?], VisibleBoundary
```
Krijovnick: There is should be `GetRowHeightFn` type instead `GetColumnWidthFn`

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
```diff
@@ -16,7 +16,7 @@ const rowsComputed = ({ virtualRows }: Getters) => plainRows(virtualRows);
 const loadedRowsStartComputed = ({ virtualRows }: Getters) => loadedRowsStart(virtualRows);
 
 // tslint:disable-next-line: max-line-length
-class VirtualTableStateBase extends React.PureComponent<VirtualTableStateProps, VirtualTableStateState> {
+class VirtualTableStateBase extends React.PureComponent<VirtualTableStateProps, any> {
```
Krijovnick: - The `VirtualTableStateState` type can be removed from `virtula-table-state.types.ts` file. 
- Should be there `any` state? Maybe `DeprecatedLazyLoadingStrategyState`?

packages/dx-react-grid/src/plugins/virtual-table/virtual-table.tsx
```diff
@@ -24,7 +34,7 @@ export const makeVirtualTable: (...args: any) => any = (Table, {
   defaultEstimatedRowHeight,
   defaultHeight,
 }) => {
-  class VirtualTable extends React.PureComponent<VirtualTableProps, VirtualTableLayoutState> {
+  class VirtualTable extends React.PureComponent<VirtualTableProps, any> {
```
Krijovnick: There is also `any` state

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
ushkal: `VirtualTableStateState` is the one we need. `DeprecatedLazyLoadingStrategyState` is from another PR :)

PR #2067, opened 2019-06-05
fix(react-grid): change processing of last page in lazy load mode

### LINKED ISSUE #2062: VirtualTable rows sometimes do not render on last page in lazy load mode
VirtualTable in Material UI rows sometimes do not render using "lazy loaded" method after you do a sort on a column.

I thought this may be an issue with my code, but I was able to reproduce[ in the online demo](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/lazy-loading/).  I admit it takes a lot of scrolling and sorting to make it happen in the demo, but happens more often in my application.

ezgif-2-360973b7fd07

In the online demo, you can make it do it reliably, by using the scrollbar and dragging to the very bottom.  Then sort one of the columns.  Rows will not render until you scroll again.  I'm hoping this is all related to the same bug.

image

ushkal:
fixes #2062

### DIFF
--- a/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
@@ -45,16 +45,19 @@ export const mergeRows: MergeRowsFn = (
 export const calculateRequestedRange: CalculateRequestedRangeFn = (
   loadedInterval, newRange, referenceIndex, pageSize,
 ) => {
-  if (Math.abs(loadedInterval.start - newRange.start) >= 2 * pageSize) {
-    const useFirstHalf = referenceIndex % pageSize < pageSize / 2;
-    const start = useFirstHalf
-      ? newRange.start
-      : newRange.start + pageSize;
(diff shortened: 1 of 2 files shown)


packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
```diff
@@ -45,16 +45,19 @@ export const mergeRows: MergeRowsFn = (
...
-    return { start, end };
+  const isAdjacentPage = Math.abs(loadedInterval.start - newRange.start) < 2 * pageSize;
+  if (isAdjacentPage) {
+    return intervalUtil.difference(newRange, loadedInterval);
   }
-  return intervalUtil.difference(newRange, loadedInterval);
+
+  const useFirstHalf = referenceIndex % pageSize < pageSize / 2;
+  const isLastPage = intervalUtil.getLength(newRange) / pageSize < 3;
+  const start = useFirstHalf || isLastPage
```
ushkal: only this condition was changed

yedukondalurd:
When can we have these changes?

PR #2069, opened 2019-06-05
docs(react-grid): add reference for createRowCache function

ushkal:


packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,30 @@
...
+Name | Type | Default | Description
+-----|------|---------|------------
+pageSize | number | 100 | The number of rows on virtual page. It must be equal to the value of pageSize property of the [VirtualTableState](virtual-table-state.md#properties) plugin.
+capacity | number | Infinity | The maximum number of pages that cache should store. When the reaches its capacity, it discards the least recently used virtual page.
+
+### Return value
+
+Type | Description
+-----|------------
+RowCache | A row cache.
```
MaximKudriavtsev: Can we add a reference to RowCache interface?
Can we add a section `How to import`?

ushkal: It's a good idea to add a usage example.
The RowCache interface described below, I think it's better to keep them in a single document

MaximKudriavtsev: No, I mean other points.

- Add [RowCache](`#RowCache`) link.
- Add import section as we did here
image

packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,38 @@
+# createRowCache
+
+A function that instantiates a default implementation of row cache. The cache is advised to be used in [lazy load mode](../guides/lazy-loading.md) to reduce the number of network requests.
```
RomanTsukanov:
```suggestion
A function that instantiates the default implementation of row caching. We recommend that you use row caching in [lazy loading mode](../guides/lazy-loading.md) to reduce the number of network requests.
```

packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,38 @@
...
+import { createRowCache } from '@devexpress/dx-react-grid';
+```
+
+## User reference
+
+### Arguments
+
+Name | Type | Default | Description
+-----|------|---------|------------
+pageSize | number | 100 | The number of rows on virtual page. It must be equal to the value of pageSize property of the [VirtualTableState](virtual-table-state.md#properties) plugin.
```
RomanTsukanov:
```suggestion
pageSize | number | 100 | The number of rows on a virtual page. It should be equal to the `pageSize` property value specified in the [VirtualTableState](virtual-table-state.md#properties) plugin.
```

packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,38 @@
...
+capacity | number | Infinity | The maximum number of pages that cache should store. When the cache reaches its capacity, it discards the least recently used virtual page.
+
+### Return value
```
RomanTsukanov:
```suggestion
### Return Value
```

packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+## User reference
+
+### Arguments
+
+Name | Type | Default | Description
+-----|------|---------|------------
+pageSize | number | 100 | The number of rows on virtual page. It must be equal to the value of pageSize property of the [VirtualTableState](virtual-table-state.md#properties) plugin.
+capacity | number | Infinity | The maximum number of pages that cache should store. When the cache reaches its capacity, it discards the least recently used virtual page.
```
RomanTsukanov:
```suggestion
capacity | number | Infinity | The maximum number of virtual pages that the cache should store. When the capacity is reached, the least recently used page is removed from the cache to make room for new pages.
```

packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Interfaces
+
+### RowCache
+
+Describes a row cache.
+
+Field | Type | Description
+------|------|------------
+getRows | (skip: number, take: number) => any[] | Returns `take` number of rows starting with the `skip` row from the cache.
```
RomanTsukanov:
```suggestion
getRows | (skip: number, take: number) => any[] | Returns from the cache the `take` number of rows starting with the `skip` row.
```

packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,38 @@
...
+setRows | (skip: number, rows: ReadonlyArray&lt;any&gt;) => void | Adds `rows` to the cache.
+invalidate | () => void | Clears the cache
```
RomanTsukanov:
```suggestion
invalidate | () => void | Clears the cache.
```

packages/dx-react-grid/docs/reference/create-row-cache.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+## User reference
+
+### Arguments
+
+Name | Type | Default | Description
+-----|------|---------|------------
+pageSize | number | 100 | The number of rows on a virtual page. It should be equal to the `pageSize` property value specified in the [VirtualTableState](virtual-table-state.md#properties) plugin.
+capacity | number | Infinity | The maximum number of virtual pages that the cache should store. When the capacity is reached, the least recently used page is removed from the cache to make room for new pages.
```
arminal:
```suggestion
capacity | number | Infinity | The maximum number of virtual pages that the cache should store. When capacity is reached, the oldest used page is removed from the cache to make room for new pages.
```

PR #2070, opened 2019-06-06
fix(react-chart): arg and val scale inconsistency

DmitryBogomolov:
Fixes https://github.com/DevExpress/devextreme-reactive/issues/2078.

Removes `fixOffset` utils - moves it into `scaleBand` wrapper.


packages/dx-chart-core/src/utils/scale.ts
```diff
@@ -11,9 +11,12 @@ import {
...
 export const scaleLinear: FactoryFn = d3ScaleLinear as any;
 /** @internal */
-export const scaleBand: FactoryFn = () => (
-  d3ScaleBand().paddingInner(0.3).paddingOuter(0.15) as any
-);
+export const scaleBand: FactoryFn = () => {
+  const scale = d3ScaleBand().paddingInner(0.3).paddingOuter(0.15);
+  const wrapper = (value: any) => scale(value)! + scale.bandwidth() / 2;
+  Object.assign(wrapper, scale);
+  return wrapper as ScaleObject;
```
MaximKudriavtsev: Why we don't use new syntax?

```js
return { ...wrapper, ...scale }
```

packages/dx-chart-core/src/utils/scale.ts
```diff
@@ -246,7 +231,24 @@ const invertBandScaleBounds = (scale: ScaleObject, range: NumberArray): DomainBo
   ];
 };
 
+// tslint:disable-next-line: ban-types
+const makeScaleHelper = <T extends Function>(choices: [T, T]) => {
+  const func: any = (scale: ScaleObject, ...args: any[]) => {
+    const choosen = choices[Number(!!scale.bandwidth)];
```
MaximKudriavtsev: Maybe we should write the next sentence more cleary? I mean something like this
```js
const choosen = scale.bandwidth === 0 ? choices[0] : choices[1]
```

packages/dx-chart-core/src/utils/scale.ts
DmitryBogomolov: That's not the same - `assign` copies all keys from `scale` to `wrapper`, spread copies all keys from `wrapper` and `scale` to a new object.

MaximKudriavtsev: But a function result is same

packages/dx-chart-core/src/utils/scale.ts
DmitryBogomolov: `scale.bandwidth` is a function (if exists) - not a number. That would be

```js
const choosen = 'bandwidth' in scale ? choices[0] : choices[1]
```

The question is - will accessing `choices` variable twice be more clear?

MaximKudriavtsev: I think so.

@Krijovnick what do you think about this?

packages/dx-chart-core/src/utils/scale.ts
DmitryBogomolov: How can it be the same?

MaximKudriavtsev: My mistake, I didn't see that wrapper is a function.

packages/dx-chart-core/src/utils/scale.ts
DmitryBogomolov: That code is no more.

Krijovnick: The last variant is more clear.

Issue #2080, opened 2019-06-10
Strange text artifacts appear during Chart panning in Edge


DmitryBogomolov:
### I'm using ...

- [ ] React Grid
- [x] React Chart
- [ ] React Scheduler

### Current Behaviour

Texts from argument axis appear under mouse cursor. Chart is not panned.

### Expected Behaviour

No texts under cursor. Chart is panned.

### Steps to Reproduce

Open `/#/demo/zoom-pan/controlled/bootstrap4/default` demo in Edge and try to pan the Chart.

### Screenshots

![chart-pan-issue](https://user-images.githubusercontent.com/2501124/59193079-39ed1880-8b8d-11e9-9966-97cba35850e6.gif)

### Environment

- devextreme-reactive: 1.11.0
- react: xxx
- browser: Edge
- bootstrap: none
- react-bootstrap: none
- material-ui: none

DmitryBogomolov:
Looks like calling [clear()](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-core/src/draggable/selection-utils.ts) on *move* handler helps.

PR #2081, opened 2019-06-11
fix(react-chart): fix strange text artifacts during scrolling

Krijovnick:
#2080


packages/dx-react-core/src/index.ts
```diff
@@ -9,6 +9,9 @@ export { TemplateConnector } from './plugin-based/template-connector';
 /** @internal */
 export { Draggable } from './draggable';
 
+/** @internal */
+export { clear } from './draggable/selection-utils';
```
ushkal: In a context of a package it's not clear what `clear` does exactly. Maybe we should rename it to `clearSelection`?

PR #2086, opened 2019-06-13
chore: add configs for support robots

churkin:


.github/lock.yml
```diff
@@ -0,0 +1,41 @@
...
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 10
+
+# Skip issues and pull requests created before a given timestamp. Timestamp must
+# follow ISO 8601 (`YYYY-MM-DD`). Set to `false` to disable
+skipCreatedBefore: false
+
+# Issues and pull requests with these labels will be ignored. Set to `[]` to disable
+exemptLabels:
```
ushkal: Shouldn't we add the next labels: 
discussion, enhancement, help wanted, more-information-needed, bug

.github/no-response.yml
```diff
@@ -0,0 +1,13 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 10
+# Label requiring a response
+responseRequiredLabel: "STATE: Need clarification"
```
ushkal: 'more-information-needed' in our case

churkin: I'll rename this label to "STATE: Need clarification". Unfortunatly this label's name is hard coded in the robot ((

.github/lock.yml
churkin: Why? If the issue is closed we can lock it.

PR #2105, opened 2019-06-20
chore: change the waiting label name

MaximKudriavtsev:
The previous name was not clear.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -10,7 +10,7 @@ skipCreatedBefore: false
 # Issues and pull requests with these labels will be ignored. Set to `[]` to disable
 exemptLabels:
    - "STATE: Need response"
-   - "STATE: Waiting for customer to reply"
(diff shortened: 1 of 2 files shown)


.github/lock.yml
```diff
@@ -10,7 +10,7 @@ skipCreatedBefore: false
 # Issues and pull requests with these labels will be ignored. Set to `[]` to disable
 exemptLabels:
    - "STATE: Need response"
-   - "STATE: Need clarification"
+   - "STATE: Waiting for customer to reply"
```
ushkal: may be 
'Awaiting customer response' or ' Waiting for customer'

MaximKudriavtsev: @RomanTsukanov what do you think about these changes?

RomanTsukanov: "Awaiting customer response" is a good alternative

PR #2106, opened 2019-06-21
fix(react-grid): calculate summary for collapsed groups under first level

LazyLahtak:


packages/dx-grid-core/src/plugins/integrated-summary/computeds.ts
```diff
@@ -97,6 +94,17 @@ export const groupSummaryValues: GroupSummaryValuesFn = (
         row,
         rows: [],
       });
+      levelIndex = levels.findIndex(level => level.levelKey === levelKey);
```
ushkal: A function should be extracted

packages/dx-grid-core/src/plugins/integrated-summary/computeds.ts
```diff
@@ -97,6 +97,17 @@ export const groupSummaryValues: GroupSummaryValuesFn = (
         row,
         rows: [],
       });
+      levelIndex = getLevelIndex(levelKey);
+    }
+    const isCollapsedNestedGroupRow = collapsedRows && levelIndex > 0;
+    if (!levelKey || isCollapsedNestedGroupRow) {
+      levels.forEach((level) => {
+        if (!levelKey) {
+          level.rows.push(row);
```
DmitryBogomolov: Why is array modified in place here and recreated in the `else` case?

```js
level.rows.push(...collapsedRows);
```

packages/dx-grid-core/src/plugins/integrated-summary/computeds.ts
```diff
@@ -97,6 +97,17 @@ export const groupSummaryValues: GroupSummaryValuesFn = (
         row,
         rows: [],
       });
+      levelIndex = getLevelIndex(levelKey);
+    }
+    const isCollapsedNestedGroupRow = collapsedRows && levelIndex > 0;
+    if (!levelKey || isCollapsedNestedGroupRow) {
+      levels.forEach((level) => {
+        if (!levelKey) {
```
DmitryBogomolov: The condition is checked on each iteration. Code can be replaced with something like this:
```js
const rowsToAppend = !levelKey ? [row] : collapsedRows;
levels.forEach((level) => {
  level.rows.push(...rowsToAppend);
});
```

packages/dx-grid-core/src/plugins/integrated-summary/computeds.ts
```diff
@@ -71,20 +71,20 @@ export const groupSummaryValues: GroupSummaryValuesFn = (
   getCellValue,
   getRowLevelKey,
   isGroupRow,
+  getCollapsedRows,
   calculator = defaultSummaryCalculator,
 ) => {
   let levels: any[] = [];
```
DmitryBogomolov: Isn't there `Level` type?

packages/dx-grid-core/src/plugins/integrated-summary/computeds.ts
ushkal: Agreed, that would be more consistent

Issue #2107, opened 2019-06-22
ZoomAndPan is not defined in the typescript definition file.


towry:
`@devexpress/dx-react-chart-material-ui: v1.11.1`

In my TS file, if I do the import:

```ts
import { ZoomAndPan } from '@devexpress/dx-react-chart-material-ui';
```
It will give me an error: `has no exported member 'ZoomAndPan'`

I have to use `require` to get the plugin to work:

```js
const ZoomAndPan = require('@devexpress/dx-react-chart-material-ui').ZoomAndPan;
```

towry:
Changed to react-vis.

DmitryBogomolov:
In [this sample](https://codesandbox.io/s/react-typescript-ssj6e) Chart works but there is an import warning with the same text.

PR #2109, opened 2019-06-24
fix(react-chart): update ZoomAndPan plugin ts declaration

DmitryBogomolov:
Fixes https://github.com/DevExpress/devextreme-reactive/issues/2107.

Renames *packages/dx-react-chart-<theme>/src/plugins/zoom-pan.jsx* files so that they are not missed by the **ts-generator**.
Renames `ViewportOptions` to `Viewport` so that it is synchronized with entities in *md* files.


packages/dx-react-chart/docs/reference/zoom-and-pan.md
```diff
@@ -23,17 +23,17 @@ import { ZoomAndPan } from '@devexpress/dx-react-chart';
...
+onViewportChange? | (viewport: [Viewport](#viewport)) => void | A function that is executed when the viewport changes.
 interactionWithArguments? | 'none' &#124; 'pan' &#124; 'zoom' &#124; 'both' | The type of interaction available for the argument scale.
 interactionWithValues? | 'none' &#124; 'pan' &#124; 'zoom' &#124; 'both' | The type of interaction available for the value scale.
 zoomRegionKey? | 'shift' &#124; 'alt' &#124; 'ctrl' | The key the allows a user to zoom a region by selecting it with the mouse drag gesture.
 dragBoxComponent? | ComponentType&lt;[ZoomAndPan.DragBoxProps](#zoomandpandragboxprops)&gt; | A component that renders the rectangle that appears when a user zooms a region.
 
 ## Interfaces
 
-### ZoomAndPan.Viewport
+### Viewport
```
Krijovnick: In most files we use the rule `Plugin_Name.Interface_Name` for declaring interfaces.

packages/dx-react-chart-material-ui/src/index.js
```diff
@@ -13,4 +13,4 @@ export { ValueAxis } from './plugins/value-axis';
 export { ArgumentAxis } from './plugins/argument-axis';
 
 export { Tooltip } from './plugins/tooltip';
-export { ZoomAndPan } from './plugins/zoom-pan';
+export { ZoomAndPan } from './plugins/zoom-and-pan';
```
Krijovnick: Should we rename the folder name `zoom-pan` in the `dx-chart-core`?

packages/dx-react-chart/docs/reference/zoom-and-pan.md
DmitryBogomolov: `PluginName` prefix affects generated ts code. The following is generated

```ts
export namespace ZoomAndPan {
  /** The viewport configuration. */
  export type Viewport = ZoomAndPanBase.Viewport;
}
```

and causes an error as there is no `Viewport` field in `ZoomAndPanBase`.

packages/dx-react-chart-material-ui/src/index.js
DmitryBogomolov: Yes, we should.

DmitryBogomolov: Renamed.

PR #2112, opened 2019-06-25
fix(react-grid): add TableGroupRow as optional dependency for TableFixedColumns

LazyLahtak:


packages/dx-react-grid/src/plugins/table-fixed-columns.tsx
```diff
@@ -39,6 +39,7 @@ const pluginDependencies = [
   { name: 'TableSelection', optional: true },
   { name: 'TableSummaryRow', optional: true },
   { name: 'TableTreeColumn', optional: true },
+  { name: 'TableGroupRow', optional: true },
```
ushkal: Let's move it between the TableFilterRow and the TableHeaderRow plugins to preserve alphabetic order

PR #2113, opened 2019-06-25
chore: make Scheduler's text unselectable

MaximKudriavtsev:


ushkal:
Should we also specify msUserSelect for ie11?

PR #2137, opened 2019-07-03
perf(react-grid): optimize render using React.memo

LazyLahtak:
Performance increase is about 10%.


churkin:
What about scheduler?

AryamnovEugeniy:
> What about scheduler?

We will deal with this in the iteration

MaximKudriavtsev:
Should we change all our `PureComponents` without state to `React.memo`?

LazyLahtak:
> Should we change all our `PureComponents` without state to `React.memo`?

It makes sense if we have many same PureComponent in app.

PR #2140, opened 2019-07-04
fix(react-scheduler): add capability to pass className and style into Scheduler layout component

AryamnovEugeniy:


packages/dx-react-scheduler-material-ui/src/templates/layouts/horizontal-view-layout.jsx
```diff
@@ -70,4 +73,9 @@ HorizontalViewLayoutBase.propTypes = {
   height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
 };
 
+
```
MaximKudriavtsev: Avoid adding excess new lines

packages/dx-react-scheduler-material-ui/src/templates/layouts/horizontal-view-layout.jsx
```diff
@@ -68,6 +73,13 @@ HorizontalViewLayoutBase.propTypes = {
   layoutRef: PropTypes.object.isRequired,
   layoutHeaderRef: PropTypes.object.isRequired,
   height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
+  className: PropTypes.string,
+  style: PropTypes.object,
+};
+
+HorizontalViewLayoutBase.defaultProps = {
+  className: undefined,
+  style: undefined,
```
ushkal: We use `null` as a default value for object props e.g. `style`

packages/dx-react-scheduler-material-ui/src/templates/layouts/horizontal-view-layout.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+    classes = getClasses(<HorizontalViewLayout {...defaultProps} />);
+    shallow = createShallow({ dive: true });
+  });
+
+  it('should pass className to the root element', () => {
+    const tree = shallow((
+      <HorizontalViewLayout {...defaultProps} className="custom-class" />
+    ));
+
+    expect(tree.find('.custom-class'))
```
ushkal: We should check root component in this tests: `tree.hasClass('custom-class')`

packages/dx-react-scheduler-material-ui/src/templates/layouts/horizontal-view-layout.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+    expect(tree.find(`.${classes.container}`))
+      .toBeTruthy();
+  });
+
+  it('should pass rest props to the root element', () => {
+    const tree = shallow((
+      <HorizontalViewLayout {...defaultProps} data={{ a: 1 }} />
+    ));
+
+    expect(tree.find(`.${classes.container}`).props().data)
```
ushkal: restProps should be applied to the root element so we should check `tree`: `tree.prop('data') ...`

ushkal: restProps should be applied to the root element so we should check `tree`: `tree.prop('data') ...`

ushkal: Same applies to other tests

PR #2141, opened 2019-07-04
chore(react-chart): change cursor type on zoom/pan

Krijovnick:


packages/dx-chart-core/src/plugins/zoom-and-pan/computeds.ts
```diff
@@ -200,6 +200,9 @@ export const getRect = (
...
 
-const checkInteraction = (interaction: Interaction, type: Interaction) => (
-  interaction === 'both' || interaction === type
-);
+const checkInteraction = (interaction: Interaction, type: Interaction) =>
+interaction === 'both' || interaction === type;
+
+/** @internal */
+export const setCursorType = (node: any, type: string) =>
+node.style.cursor = type;
```
MaximKudriavtsev: Why we return a value here? This function is not pure, I think we should return `void`.

packages/dx-react-chart/src/plugins/zoom-and-pan.tsx
```diff
@@ -34,21 +35,25 @@ const events = {
 
 class ZoomPanProvider extends React.PureComponent<ZoomPanProviderProps> {
   handlers!: EventHandlers;
-  ref!: Element;
+  ref!: SVGElement;
   windowHandlers!: { [key: string]: EventHandlers};
 
   componentDidMount() {
-    this.ref = this.props.rootRef.current!;
+    this.ref = this.props.rootRef.current! as SVGElement;
```
MaximKudriavtsev: The `ref` variable name is not clear to understand.

PR #2145, opened 2019-07-05
fix(react-scheduler): set px as measurement unit of MonthView today cell lineHeight

AryamnovEugeniy:


packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/time-table/cell.jsx
```diff
@@ -36,7 +36,7 @@ const styles = theme => ({
     display: 'inline-block',
     width: theme.spacing(3),
     height: theme.spacing(3),
-    lineHeight: theme.spacing(3),
+    lineHeight: `${theme.spacing(3)}px`,
```
ushkal: Shouldn't the size of this element depend on typography rather than spacing?

PR #2153, opened 2019-07-09
fix(react-grid): disable skeleton cells when local data is used

LazyLahtak:


packages/dx-react-grid/src/plugins/virtual-table/virtual-table.tsx
```diff
@@ -107,7 +107,14 @@ export const makeVirtualTable: (...args: any) => any = (Table, {
             predicate={({ tableRow }: any) => !!isStubTableCell(tableRow)}
           >
             {(params: TableNS.CellProps) => (
-              <SkeletonStubCell {...params} />
+              <TemplateConnector>
+                {({ isDataRemote }) => {
+                  if (isDataRemote) {
```
ushkal: It's better to use ternary operator here

packages/dx-react-grid/src/plugins/virtual-table/virtual-table.test.tsx
```diff
@@ -218,6 +221,39 @@ describe('#makeVirtualTable', () => {
       .toMatchObject(deps.template.tableCell);
   });
 
+  it('should not render skeleton stub cell when data local', () => {
```
ushkal: ... when data **is** local

PR #2156, opened 2019-07-10
refactor(react-scheduler-material-ui): refactor styles measurement units

AryamnovEugeniy:


packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/time-table/cell.jsx
```diff
@@ -29,14 +29,15 @@ const styles = theme => ({
...
+    padding: `${theme.spacing(1)}px`,
   },
   today: {
-    margin: theme.spacing(0.5),
+    margin: `${theme.spacing(0.5)}px`,
     display: 'inline-block',
-    width: theme.spacing(3),
-    height: theme.spacing(3),
-    lineHeight: `${theme.spacing(3)}px`,
+    // TODO
```
MaximKudriavtsev: Should be removed

packages/dx-react-scheduler-material-ui/src/templates/all-day-panel/cell.jsx
```diff
@@ -9,7 +9,7 @@ import { getBorder } from '../utils';
 const styles = theme => ({
   cell: {
     padding: 0,
-    height: theme.spacing(7),
+    height: `${theme.spacing(7)}px`,
```
ushkal: A customer can customize spacing using other units such as rem. Will it work?
https://material-ui.com/customization/spacing/

MaximKudriavtsev: @ushkal you are right. We should revert to the old code.

PR #2159, opened 2019-07-10
fix(react-chart): improve touch events on hover/select

Krijovnick:


MaximKudriavtsev:
I suggest some changes

- Can we combine old behavior with the new? I mean that `touchstart` + `touchmove`.
- How a customer can hide a hover on a mobile device?
- I suggest renaming this PR. We should use a verb in the first place that describes changes type. In my opinion, `touch events on hover/select` is not clear to understand.
- Why we use `chore` in this title? The changes change behavior and our customers should know about it from our CHANGELOG.md

Krijovnick:
1. We can do that, but It is interesting question: Will user move finger on screen to change the hover/selection? I think no.
2. Customer can touch the empty area to hide a hover.
3.  and 4. Ok, I will change it.

PR #2169, opened 2019-07-16
perf(react-scheduler): optimize render using React.memo 

AryamnovEugeniy:


packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/day-scale/cell.jsx
```diff
@@ -22,7 +22,7 @@ const styles = theme => ({
   },
 });
 
-const CellBase = ({
+const CellBase = React.memo(({
```
MaximKudriavtsev: Should we use `React.memo` here? I think it is not a good idea, because we have a lot of cells on views.

And same with other `cells`

packages/dx-react-scheduler/src/plugins/scheduler-core.tsx
```diff
@@ -4,6 +4,9 @@ import {
 } from '@devexpress/dx-react-core';
 import { appointments, formatDateTimeGetter } from '@devexpress/dx-scheduler-core';
 import { SchedulerProps } from '../types';
+import { memoize } from '@devexpress/dx-core';
+
+const memoFormatDateTimeGetter = memoize(locale => formatDateTimeGetter(locale));
```
MaximKudriavtsev: Should we use `memo` here?

packages/dx-react-scheduler-material-ui/src/templates/views/horizontal/day-scale/cell.jsx
AryamnovEugeniy: These (and other) cells are rendered every time an appointment is changed, added or deleted. And they shouldn't

packages/dx-react-scheduler/src/plugins/scheduler-core.tsx
AryamnovEugeniy: We need to use `memo` here in order to be able to memoize cells.

packages/dx-react-scheduler/src/plugins/all-day-panel.tsx
```diff
@@ -59,6 +59,9 @@ class AllDayPanelBase extends React.PureComponent<AllDayPanelProps, AllDayPanelS
     this.setState({ rects, elementsMeta: cellElementsMeta });
   });
 
+  getMemoMessageFormatter = memoize(messages =>
+    getMessagesFormatter({ ...defaultMessages, ...messages }));
+
```
MaximKudriavtsev: Can we extract this function to `core` package? I think we can use it in all of our packages.

AryamnovEugeniy: Yes, we can

packages/dx-react-scheduler/src/plugins/view-switcher.tsx
```diff
@@ -17,6 +18,9 @@ class ViewSwitcherBase extends React.PureComponent<ViewSwitcherProps> {
   static components: PluginComponents = {
     switcherComponent: 'Switcher',
   };
+
+  getMemoCurrentView = memoize((name, type, displayName) => ({ name, type, displayName }));
```
MaximKudriavtsev: We should extract this code outside the class.

Also, can we extract new global constants and function into `scheduler-core` package?

packages/dx-react-scheduler/src/plugins/date-navigator.tsx
```diff
@@ -24,6 +25,8 @@ const navigate = (action, currentView, intervalCount) => (direction, nextDate) =
   amount: intervalCount,
   step: currentView.type,
 });
+const navigateMemo = memoize((changeCurrentDate, currentView, intervalCount) =>
+  navigate(changeCurrentDate, currentView, intervalCount));
```
MaximKudriavtsev: We should use only [pure functions](https://en.wikipedia.org/wiki/Pure_function) in our code.

packages/dx-react-scheduler/src/plugins/scheduler-core.tsx
MaximKudriavtsev: Can we change the `formatDateTimeGetter` function code without creating a new function? I mean, that we should use memo inside `scheduler-core` package.

packages/dx-react-scheduler/src/plugins/date-navigator.tsx
LazyLahtak: React.useCallback can be used in functional component, not in classes. If we want use it, needs to refactor component.

packages/dx-react-scheduler-material-ui/src/templates/date-navigator/root.jsx
```diff
@@ -12,6 +13,9 @@ const styles = ({ spacing }) => ({
   },
 });
 
+const onNavigateBackMemo = memoize(onNavigate => () => onNavigate('back'));
+const onNavigateForwardMemo = memoize(onNavigate => () => onNavigate('forward'));
```
MaximKudriavtsev: We should not use `memoize` in the global context

packages/dx-react-scheduler/src/plugins/all-day-panel.tsx
```diff
@@ -70,7 +73,7 @@ class AllDayPanelBase extends React.PureComponent<AllDayPanelProps, AllDayPanelS
       messages,
     } = this.props;
     const { rects, elementsMeta } = this.state;
-    const getMessage = getMessagesFormatter({ ...defaultMessages, ...messages });
+    const getMessage = this.getMemoMessageFormatter(messages);
```
MaximKudriavtsev: We should use only pure functions.

`this.getMemoMessageFormatter(messages, defaultMessages)`

packages/dx-react-scheduler/src/plugins/date-navigator.tsx
MaximKudriavtsev: We should not use `memoize` in the global contex

packages/dx-react-scheduler/src/plugins/view-switcher.tsx
```diff
@@ -35,7 +39,8 @@ class ViewSwitcherBase extends React.PureComponent<ViewSwitcherProps> {
               setCurrentViewName,
             }) => (
               <Switcher
-                currentView={currentView}
+                currentView={this.getMemoCurrentView(currentView.name,
```
MaximKudriavtsev: Should we use this function here?

AryamnovEugeniy: We need to use this function because ViewSwitcher renders every time a date is changed. We need to use the function in order to avoid unnecessary renders

AryamnovEugeniy:
FPR

ushkal:
`getAppointmentStyle` should be memoized too

MaximKudriavtsev:
@ushkal 

I think we can't do it because of this function call inside the `map` function and its results are not equals

PR #2175, opened 2019-07-18
chore: publish 2.0.1

churkin:


packages/dx-react-grid/package.json
```diff
@@ -74,7 +74,7 @@
     "typescript": "^3.5.3"
   },
   "dependencies": {
-    "@devexpress/dx-grid-core": "2.0.0"
+    "@devexpress/dx-grid-core": "2.0.1"
   },
   "peerDependencies": {
     "@devexpress/dx-core": "2.0.0",
```
MaximKudriavtsev: We should update our `peerDependencies` also

PR #2184, opened 2019-07-22
refactor(react-grid-demos): rework data fetch in lazy loading demo

LazyLahtak:


packages/dx-react-grid-demos/src/demo-sources/grid-lazy-loading/remote-data.jsxt
```diff
@@ -42,100 +42,48 @@ const DateTypeProvider = props => (
...
+  ]);
+  const [skip, setSkip] = useState(0);
+  const [requestedSkip, setRequestedSkip] = useState(0);
+  const [take, setTake] = useState(VIRTUAL_PAGE_SIZE * 2);
+  const [totalCount, setTotalCount] = useState(MAX_ROWS);
+  const [loading, setLoading] = useState(false);
+  const [forceReload, setForceReload] = useState(false);
+  const [filters, setFilters] = useState([]);
+  const [sorting, setSorting] = useState([]);
+  const [lastQuery, setLastQuery] = useState();
```
ushkal: `lastQuery` should belong to instance not state

packages/dx-react-grid-demos/src/demo-sources/grid-lazy-loading/remote-data.jsxt
```diff
@@ -18,6 +18,7 @@ const VIRTUAL_PAGE_SIZE = 100;
 const MAX_ROWS = 50000;
 const URL = 'https://js.devexpress.com/Demos/WidgetsGalleryDataService/api/Sales';
 const getRowId = row => row.Id;
+let lastQuery;
```
ushkal: It should be declared inside function

LazyLahtak: If we will do it, every render it will be `undefined` (and we catch infnite loading).
We should declare it outside or use `useState`

Issue #2114, opened 2019-06-25
Data not loading when sorting in already cached data chunk


michaelfaisst:
- [x] I have searched this repository's [issues](https://github.com/devexpress/devextreme-reactive/issues) and believe that this is not a duplicate.

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

When changing the sorting of a lazy-loaded virtual table, while beeing in an area of the grid which has been chached, the `getRows` method will not be called.

### Expected Behaviour

`getRows` should be called to fetch the data with the new sorting.

### Steps to Reproduce

You can reproduce this step in your "Row caching" example [here](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/lazy-loading/).

1. Open your browsers dev tools and switch to network request
2. Scroll down the table (after around row 105 will be visible) until the second data request will be sent (should be skip=200, take=100)
3. Now scroll up to the top of grid again
4. Sort any column

After following these steps data will not be loaded and the "gray box" loading state will be displayed permanently. I'm guessing that it has something to do with `skip` beeing to high for the currently displayed range of data.

![image](https://user-images.githubusercontent.com/1477993/60102772-a3a91d00-975e-11e9-8d2a-32814f84f760.png)

### Environment

- devextreme-reactive: 1.11.1
- react: 16.8.2

LazyLahtak:
Hi @michaelfaisst,
 
We have fixed this bug. The fix will be available in the next update.

PR #2186, opened 2019-07-23
fix(react-grid): correctly calculate rows to load when sorting/filtering in lazy loading mode 

LazyLahtak:
Fixes #[2114](https://github.com/DevExpress/devextreme-reactive/issues/2114)


packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
```diff
@@ -34,15 +34,18 @@ class VirtualTableStateBase extends React.PureComponent<VirtualTableStateProps,
...
+    { virtualRows }: Getters,
   ) => {
     const { pageSize, totalRowCount } = this.props;
 
     let newBounds;
     let requestedRange;
     let actualVirtualRows = virtualRows;
     if (forceReload) {
-      newBounds = requestedRange = { start: skip, end: skip + pageSize! * 2 };
+      const loadedInterval = intervalUtil.getRowsInterval(virtualRows);
```
ushkal: loadedInterval is declared twice

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
```diff
@@ -34,18 +34,22 @@ class VirtualTableStateBase extends React.PureComponent<VirtualTableStateProps,
...
     const { pageSize, totalRowCount } = this.props;
 
     let newBounds;
     let requestedRange;
     let actualVirtualRows = virtualRows;
+    let loadedInterval;
     if (forceReload) {
-      newBounds = requestedRange = { start: skip, end: skip + pageSize! * 2 };
+      loadedInterval = intervalUtil.getRowsInterval(virtualRows);
+      newBounds = requestedRange = getForceReloadInterval(
```
MaximKudriavtsev: Why we are using this notation `a = b = value` and are not using `let a, b, c;`?

I think these things are the same. We should not use `a = b = value`

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
```diff
@@ -326,4 +326,20 @@ describe('VirtualTableState helpers', () => {
       });
     });
   });
+
+  describe('#getForceReloadInterval', () => {
+    it('should return loadedInterval.start + pageSize * 2 as end', () => {
```
ushkal: ... if loaded interval is less than 2 pages

ushkal: should return 2 pages if loaded interval is less than 2 pages

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
```diff
@@ -326,4 +326,20 @@ describe('VirtualTableState helpers', () => {
...
+      expect(getForceReloadInterval({ start: 100, end: 200 }, 100, 1000)).toEqual({
+        start: 100,
+        end: 300,
+      });
+    });
+
+    it('should return loadedInterval.end as end', () => {
```
ushkal: should return loaded interval if it is more than 2 pages

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
```diff
@@ -34,15 +34,18 @@ class VirtualTableStateBase extends React.PureComponent<VirtualTableStateProps,
...
   ) => {
     const { pageSize, totalRowCount } = this.props;
 
     let newBounds;
     let requestedRange;
     let actualVirtualRows = virtualRows;
     if (forceReload) {
-      newBounds = requestedRange = { start: skip, end: skip + pageSize! * 2 };
+      const loadedInterval = intervalUtil.getRowsInterval(virtualRows);
+      newBounds = requestedRange = getDoublePageInterval(
```
ushkal: loaded interval should be calculated in a single place

LazyLahtak: Done

PR #2188, opened 2019-07-24
fix(react-grid): correct calculation of available row count in infninite scrolling mode

### LINKED ISSUE #2197: An empty space appears after the end of the table on infinite scrolling
When you reach the end of the table on infinite scrolling, an empty space appears. This looks like new rows are loading. The size of the space equals the pageSize property of the VirtualTableState plugin.

image
image

LazyLahtak:
Fix #2197

### DIFF
--- a/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
@@ -328,11 +328,12 @@ describe('VirtualTableState helpers', () => {
   });
 
   describe('#getAvailableRowCount', () => {
-    it('should return totalCount when not infiniteScroll', () => {
+    const nextRowCount = 200;
+    const lastRowCount = 100;
+    const totalRowCount = 1000;
+
+    it('should return totalCount when not infinite scrolling', () => {
       const isInfniniteScroll = false;
-      const nextRowCount = 200;
-      const lastRowCount = 100;
-      const totalRowCount = 1000;
 
       expect(getAvailableRowCount(
         isInfniniteScroll,
@@ -342,11 +343,8 @@ describe('VirtualTableState helpers', () => {
       )).toEqual(totalRowCount);
     });
 
-    it('should return max of nextRowCount or lastRowCount when infiniteScroll', () => {
+    it('should return max of nextRowCount or lastRowCount when infinite scrolling', () => {
       const isInfniniteScroll = true;
-      const nextRowCount = 200;
-      const lastRowCount = 100;
-      const totalRowCount = 1000;
 
       expect(getAvailableRowCount(
         isInfniniteScroll,
--- a/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
@@ -368,6 +368,8 @@ describe('VirtualTableState helpers', () => {
         lastRowCount,
         totalRowCount,
       )).toEqual(lastRowCount);
+    });
+  });
 
   describe('#getForceReloadInterval', () => {
     it('should return 2 pages if loaded interval is less than 2 pages', () => {
--- a/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
@@ -328,7 +328,7 @@ describe('VirtualTableState helpers', () => {
   });
 
   describe('#getAvailableRowCount', () => {
-    const nextRowCount = 200;
+    const newRowCount = 200;
     const lastRowCount = 100;
     const totalRowCount = 1000;
 
@@ -337,21 +337,21 @@ describe('VirtualTableState helpers', () => {
 
       expect(getAvailableRowCount(
         isInfniniteScroll,
-        nextRowCount,
+        newRowCount,
         lastRowCount,
         totalRowCount,
       )).toEqual(totalRowCount);
     });
 
-    it('should return max of nextRowCount or lastRowCount when infinite scrolling', () => {
+    it('should return max of newRowCount or lastRowCount when infinite scrolling', () => {
       const isInfniniteScroll = true;
 
       expect(getAvailableRowCount(
         isInfniniteScroll,
-        nextRowCount,
+        newRowCount,
         lastRowCount,
         totalRowCount,
-      )).toEqual(Math.max(nextRowCount, lastRowCount));
+      )).toEqual(Math.max(newRowCount, lastRowCount));
     });
 
   });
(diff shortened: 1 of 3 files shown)


packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
```diff
@@ -101,8 +101,12 @@ export const trimRowsToInterval: PureComputed<[VirtualRows, Interval]> = (
...
 export const getAvailableRowCount: PureComputed<[boolean, number, number, number], number> = (
   infiniteScroll, newCount, lastCount, totalRowCount,
-) => (
-  infiniteScroll
-    ? Math.max(newCount, lastCount)
+) => {
+  const maxWhenInfiniteScrolling = Math.min(
+    Math.max(newCount, lastCount),
+    totalRowCount);
+  return (infiniteScroll
```
MaximKudriavtsev: If `infiniteScroll` is a boolean variable, I suggest using the next variable name - `isInfiniteScrolling`

MaximKudriavtsev: You have 3 variables - `newCount`, `lastCount`, `totalRowCount`. You should add the `Row` word into each name or remove this word from `totalRowCount`

MaximKudriavtsev: You calculate `maxWhenInfiniteScrolling` variable always but use it only `infiniteScroll` is `true`

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
```diff
@@ -75,9 +75,12 @@ class VirtualTableStateBase extends React.PureComponent<VirtualTableStateProps,
       getRows(newPageIndex, loadCount);
 
       const virtualRowsCache = trimRowsToInterval(virtualRows, newBounds);
-      const newRowCount = infiniteScrolling
-        ? Math.max(newBounds.end + loadCount, stateAvailableCount)
-        : totalRowCount;
+      const newRowCount = getAvailableRowCount(
+        infiniteScrolling,
```
MaximKudriavtsev: `infiniteScrolling` -> `isInfiniteScrolling` ?

MaximKudriavtsev: newRowCount -> nextRowCount ?
But we should use only one word for variables. `next` / `new`

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
MaximKudriavtsev: We should add tests for new functionality

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
LazyLahtak: This `infiniteScrolling` comes from props. 
We can do: `{..., infiniteScrolling: isInfiniteScrolling, ... } = this.props`, but should we?

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
LazyLahtak: > You calculate `maxWhenInfiniteScrolling` variable always but use it only `infiniteScroll` is `true`

I can remove `maxWhenInfiniteScrolling` variable and calculate this in expression. But it may do code less readable. Should we do that?

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
MaximKudriavtsev: @ushkal,

Maybe we should change the prop name too?

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
```diff
@@ -326,4 +326,35 @@ describe('VirtualTableState helpers', () => {
...
+        lastRowCount,
+        totalRowCount,
+      )).toEqual(totalRowCount);
+    });
+
+    it('should return max of nextRowCount or lastRowCount when infiniteScroll', () => {
+      const isInfniniteScroll = true;
+      const nextRowCount = 200;
+      const lastRowCount = 100;
+      const totalRowCount = 1000;
```
MaximKudriavtsev: We can extract these variables to `describe` section to avoid copy paste

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
```diff
@@ -326,4 +326,35 @@ describe('VirtualTableState helpers', () => {
       });
     });
   });
+
+  describe('#getAvailableRowCount', () => {
+    it('should return totalCount when not infiniteScroll', () => {
```
MaximKudriavtsev: `should return totalCount when not infiniteScroll`
->
`should return totalCount when not infinite scrolling`

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
MaximKudriavtsev: > We can do: {..., infiniteScrolling: isInfiniteScrolling, ... } = this.props, but should we?

I think we should

ushkal: prop name was coordinated with techwriters and I don't see why the current name is bad. Besides that, renaming prop is a bc

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
```diff
@@ -326,4 +326,33 @@ describe('VirtualTableState helpers', () => {
...
+
+    it('should return max of newRowCount or lastRowCount when infinite scrolling', () => {
+      const isInfniniteScroll = true;
+
+      expect(getAvailableRowCount(
+        isInfniniteScroll,
+        newRowCount,
+        lastRowCount,
+        totalRowCount,
+      )).toEqual(Math.max(newRowCount, lastRowCount));
```
ushkal: We should be split to 2 tests

LazyLahtak:
FPR

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
ushkal: It will be more readable with implicit return

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
```diff
@@ -327,6 +328,49 @@ describe('VirtualTableState helpers', () => {
...
+
+      expect(getAvailableRowCount(
+        isInfniniteScroll,
+        newRowCount,
+        lastRowCount,
+        totalRowCount,
+      )).toEqual(totalRowCount);
+    });
+
+    it('should return newRowCount if it more than lastRowCount in infinite scrolling', () => {
```
ushkal: Let's extract these 2 tests to a describe block named 'Infinite scrolling mode'

PR #2191, opened 2019-07-24
fix(react-chart): fix tooltip customization

### LINKED ISSUE #2166: React Chart - allow Tooltip customization
### I'm using ...

- [ ] React Grid
- [x] React Chart
- [ ] React Scheduler

### Description

I'm not able to customize styles of Tooltip (specifically changing the background color), it does provide OverlayComponent but it's the top level and the styles are overridden by MuiPaper, there is also ContentComponent, but it's the actual content of the mui paper. I want to be able to change background color of the whole tooltip, including MuiPaper div as well as the arrow.

Krijovnick:
Fixes #2166 

Added `arrowComponent` and `sheetComponent` to customize background of the tooltip and its arrow.

### DIFF
--- a/packages/dx-react-bootstrap4/components/popover.jsx
+++ b/packages/dx-react-bootstrap4/components/popover.jsx
@@ -6,7 +6,7 @@ import * as PropTypes from 'prop-types';
 import { Popper } from 'react-popper';
 
 const DefaultArrowComponent = ({ arrowProps }) => (
-  <div className="arrow" ref={arrowProps.ref} style={arrowProps.style} />
+  <div className="arrow" {...arrowProps} />
 );
 
 DefaultArrowComponent.propTypes = {
--- a/packages/dx-react-bootstrap4/components/popover.jsx
+++ b/packages/dx-react-bootstrap4/components/popover.jsx
@@ -78,9 +78,9 @@ export class Popover extends React.PureComponent {
   renderPopper() {
     const {
       children, target, renderInBody,
-      arrowComponent: ArrowComponent = DefaultArrowComponent, ...restProps
+      arrowComponent: ArrowComponent, ...restProps
     } = this.props;
-
+    const { placement: poperPlacement } = this.props;
     return (
       <Popper
         referenceElement={target}
@@ -93,7 +93,7 @@ export class Popover extends React.PureComponent {
             <div className="popover-inner" ref={this.contentRef}>
               {children}
             </div>
-            <ArrowComponent arrowProps={arrowProps} placement={placement} />
+            <ArrowComponent arrowProps={arrowProps} placement={poperPlacement} />
           </div>
         )}
       </Popper>
@@ -137,5 +137,5 @@ Popover.defaultProps = {
   isOpen: false,
   placement: 'auto',
   toggle: undefined,
-  arrowComponent: undefined,
+  arrowComponent: DefaultArrowComponent,
 };
--- a/packages/dx-react-bootstrap4/components/popover.jsx
+++ b/packages/dx-react-bootstrap4/components/popover.jsx
@@ -5,6 +5,14 @@ import * as ReactDOM from 'react-dom';
 import * as PropTypes from 'prop-types';
 import { Popper } from 'react-popper';
 
+const DefaultArrowComponent = ({ arrowProps }) => (
+  <div className="arrow" ref={arrowProps.ref} style={arrowProps.style} />
+);
+
+DefaultArrowComponent.propTypes = {
+  arrowProps: PropTypes.object.isRequired,
+};
+
 export class Popover extends React.PureComponent {
   constructor(props) {
     super(props);
@@ -69,7 +77,8 @@ export class Popover extends React.PureComponent {
 
   renderPopper() {
     const {
-      children, target, renderInBody, ...restProps
+      children, target, renderInBody,
+      arrowComponent: ArrowComponent = DefaultArrowComponent, ...restProps
     } = this.props;
 
     return (
@@ -84,7 +93,7 @@ export class Popover extends React.PureComponent {
             <div className="popover-inner" ref={this.contentRef}>
               {children}
             </div>
-            <div className="arrow" ref={arrowProps.ref} style={arrowProps.style} />
+            <ArrowComponent arrowProps={arrowProps} placement={placement} />
           </div>
         )}
       </Popper>
@@ -119,6 +128,7 @@ Popover.propTypes = {
     PropTypes.object,
   ]),
   toggle: PropTypes.func,
+  arrowComponent: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
 };
 
 Popover.defaultProps = {
@@ -127,4 +137,5 @@ Popover.defaultProps = {
   isOpen: false,
   placement: 'auto',
   toggle: undefined,
+  arrowComponent: undefined,
 };
--- a/packages/dx-react-bootstrap4/components/popover.jsx
+++ b/packages/dx-react-bootstrap4/components/popover.jsx
@@ -5,14 +5,14 @@ import * as ReactDOM from 'react-dom';
 import * as PropTypes from 'prop-types';
 import { Popper } from 'react-popper';
 
-const DefaultArrowComponent = ({ placement, refEl, ...restProps }) => (
-  <div className="arrow" ref={refEl} {...restProps} />
+const Arrow = ({ placement, ...restProps }, ref) => (
+  <div className="arrow" ref={ref} {...restProps} />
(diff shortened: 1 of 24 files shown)


packages/dx-react-chart-bootstrap4/src/templates/tooltip/sheet.test.jsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import { Sheet } from './sheet';
+
+describe('Content', () => {
```
MaximKudriavtsev: Content -> Sheet

packages/dx-react-chart-bootstrap4/src/templates/tooltip/sheet.test.jsx
```diff
@@ -0,0 +1,20 @@
...
+  const defaultProps = {
+    className: 'custom_class_name',
+    placement: 'placement',
+  };
```
MaximKudriavtsev: These properties are not required. I think we should remove `defaultProps` variable.

The next tests have the same issue.

packages/dx-react-chart-bootstrap4/src/templates/tooltip/arrow.test.jsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import { Arrow } from './arrow';
+
+describe('Content', () => {
```
MaximKudriavtsev: Content -> Arrow

packages/dx-react-chart-bootstrap4/src/templates/tooltip/arrow.jsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export const Arrow = ({ arrowProps, className }) => (
+  <div className={classNames('arrow', className)} {...arrowProps} />
+);
+Arrow.propTypes = {
+  arrowProps: PropTypes.object.isRequired,
```
MaximKudriavtsev: This property is not documented. Should we do that?

- This component doesn't spread rest properties.
- According to the documentation, this component has property `placement`.

packages/dx-react-bootstrap4/components/popover.jsx
```diff
@@ -127,4 +137,5 @@ Popover.defaultProps = {
   isOpen: false,
   placement: 'auto',
   toggle: undefined,
+  arrowComponent: undefined,
```
MaximKudriavtsev: Why the default prop for `arrowComponent` is `undefined`?

Maybe we should use a function `() => null`?

Or the next syntax for default prop `arrowComponent: DefaultArrowomponent`

packages/dx-react-bootstrap4/components/popover.jsx
```diff
@@ -69,7 +77,8 @@ export class Popover extends React.PureComponent {
 
   renderPopper() {
     const {
-      children, target, renderInBody, ...restProps
+      children, target, renderInBody,
+      arrowComponent: ArrowComponent = DefaultArrowComponent, ...restProps
```
MaximKudriavtsev: And remove this `ArrowComponent = DefaultArrowComponent`

packages/dx-react-chart/docs/reference/tooltip.md
```diff
@@ -26,6 +26,8 @@ targetItem? | [SeriesRef](./event-tracker.md#seriesref) | | An item for which th
 onTargetItemChange? | (target: [SeriesRef](./event-tracker.md#seriesref)) => void | | A function that is executed when the target item changes.
 overlayComponent | ComponentType&lt;[Tooltip.OverlayProps](#tooltipoverlayprops)&gt; | | A component that renders the tooltip.
```
MaximKudriavtsev: Overlay component receives an `arrowComponent` property.

packages/dx-react-chart-bootstrap4/src/templates/tooltip/overlay.jsx
```diff
@@ -20,9 +20,7 @@ export class Overlay extends React.PureComponent {
         modifiers={popperModifiers}
         {...restProps}
       >
-        <div className="popover-body">
-          {children}
-        </div>
+        {children}
```
MaximKudriavtsev: bs4 and mui files are not the same. Where is `arrowComponent` here? and it missing in the prop validation

Krijovnick: The `arrowComponent` in the popover.jsx

packages/dx-react-chart-bootstrap4/src/templates/tooltip/arrow.jsx
Krijovnick: - The `arrowProps` property is used due to features of `react-popper`. I think it should not be documented.
- I add `placement` property.
- I add rest properties.

packages/dx-react-bootstrap4/components/popover.jsx
```diff
@@ -5,6 +5,14 @@ import * as ReactDOM from 'react-dom';
 import * as PropTypes from 'prop-types';
 import { Popper } from 'react-popper';
 
+const DefaultArrowComponent = ({ arrowProps }) => (
+  <div className="arrow" ref={arrowProps.ref} style={arrowProps.style} />
```
churkin: `<div className="arrow" {...arrowProps} />`

packages/dx-react-bootstrap4/components/popover.jsx
```diff
@@ -84,7 +92,7 @@ export class Popover extends React.PureComponent {
             <div className="popover-inner" ref={this.contentRef}>
               {children}
             </div>
-            <div className="arrow" ref={arrowProps.ref} style={arrowProps.style} />
+            <ArrowComponent arrowProps={arrowProps} placement={restProps.placement} />
```
churkin: I don't like `arrowProps` name

packages/dx-react-chart-bootstrap4/src/templates/tooltip/arrow.jsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export const Arrow = ({
+  arrowProps, className, placement, ...restProps
+}) => (
+  <div className={classNames('arrow', className)} {...arrowProps} {...restProps} />
```
churkin:
```jsx
export const Arrow = ({
  className, ...restProps
}) => (
  <div className={classNames('arrow', className)} {...restProps} />
```

Krijovnick: It is not working. The `arrowProps` should be spread.

packages/dx-react-bootstrap4/components/popover.jsx
Krijovnick: Hm, I have no idea how to name `arrowProps` property. Maybe @MaximKudriavtsev and @ushkal can prompt?

ushkal: I think we can use `arrowProps` like in react-popper itself: https://github.com/FezVrasta/react-popper#children. It will be easier for user to understand this code

churkin: `Popper.children.arrowProps` vs `DefaultArrowComponent. arrowProps`.
Does `ArrowComponent` have other props than `arrowProps`?

packages/dx-react-chart-bootstrap4/src/templates/tooltip/arrow.jsx
churkin: Should we ignore `placement`?

Krijovnick: We can, because of `react-popper` filter the unnecessary properties. But what if it stop filter?

Krijovnick:
FPR

packages/dx-react-chart-bootstrap4/src/templates/tooltip/arrow.jsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export const Arrow = ({
+  arrowProps, className, placement, ...restProps
```
ushkal: We can use forwardRef here

PR #2208, opened 2019-07-31
chore: update deps

Krijovnick:


packages/dx-react-chart-bootstrap4/package.json
```diff
@@ -60,14 +60,14 @@
...
     "eslint-plugin-filenames": "^1.3.2",
-    "eslint-plugin-import": "^2.18.0",
-    "eslint-plugin-jest": "^22.10.0",
+    "eslint-plugin-import": "^2.18.2",
+    "eslint-plugin-jest": "^22.14.1",
     "eslint-plugin-jsx-a11y": "^6.2.3",
-    "eslint-plugin-react": "^7.14.2",
+    "eslint-plugin-react": "^7.14.3",
     "jest": "^24.8.0",
     "react": "^16.8.3",
```
ushkal: Please update react to 16.8.6

MaximKudriavtsev:
Looks like new `material-ui-pickers` version has a [issue](https://github.com/mui-org/material-ui-pickers/issues/1219) with jest tests. I think we should use a stable version of this library. I downgraded it.

PR #2219, opened 2019-08-05
chore: check peer dependency versions before publishing

churkin:


scripts/check-peer-deps.js
```diff
@@ -0,0 +1,34 @@
+const { join } = require('path');
+const {
+  existsSync,
+  readdirSync,
+  readFileSync,
+  writeFileSync
+} = require('fs');
+
+module.exports = () => {
+  console.log('Check internal peer dependency versions...');
```
ushkal: Let's instead update the versions automatically

Issue #2204, opened 2019-07-30
React Chart - points shouldn't cliped


Krijovnick:
### Description
- Points must be hiding if their center is outside the container. 
- Remove `<g>` with clip rect
- Add clip rect on series
- filter grids like ticks

PR #2223, opened 2019-08-07
chore(react-chart): prevent points clipping

Krijovnick:
#2204


packages/dx-chart-core/src/plugins/axis/computeds.ts
```diff
@@ -105,37 +102,30 @@ export const axisCoordinates: AxisCoordinatesFn = ({
...
 /** @internal */
-export const createTickFilter = ([width, height]: NumberArray) => (
-  width > 0
-    ? (tick: Tick) => tick.x1 >= 0 && tick.x1 <= width
-    : (tick: Tick) => tick.y1 >= 0 && tick.y1 <= height
-);
-
-const horizontalGridOptions = { y: 0, dy: 1 };
-const verticalGridOptions = { x: 0, dx: 1 };
+export const gridCoordinates: TickCoordinatesFn<Grid> = ({ isHor }) => {
```
ushkal: Seems that it should be `getGridCoordinates` or `gridCoordinatesGetter`?

Krijovnick: I'll rename method

PR #2232, opened 2019-08-13
fix(react-chart): fix arrow displaying on corner point

Krijovnick:
Fixed only in MUI, in the bootstrap it is ok.
before:

after:


packages/dx-react-chart-material-ui/src/templates/tooltip/arrow.jsx
```diff
@@ -50,8 +47,21 @@ const styles = (theme) => {
...
+
+BaseArrow.propTypes = {
+  placement: PropTypes.string.isRequired,
+  className: PropTypes.string,
+  classes: PropTypes.object,
+};
+
+BaseArrow.defaultProps = {
+  className: undefined,
+  classes: undefined,
```
MaximKudriavtsev: We often mark the `classes` property as a required

packages/dx-react-chart-material-ui/src/templates/tooltip/overlay.jsx
```diff
@@ -18,25 +19,46 @@ const styles = (theme) => {
...
+  arrow: {
+    element: arrowRef,
+  },
+});
 
-export const Overlay = withStyles(styles)(({
+const OverlayBase = ({
   classes, className, children, target, rotated, arrowComponent: ArrowComponent, ...restProps
 }) => {
+  const [arrowRef, setArrowRef] = React.useState(null);
```
ushkal: `useRef` is better for this task

ushkal: I meant React.useRef hook :)

Krijovnick: If we use `useRef`, then we should use `useEffect` and it is not looks better.

packages/dx-react-chart-material-ui/src/templates/tooltip/overlay.jsx
```diff
@@ -18,25 +19,46 @@ const styles = (theme) => {
...
+  classes: PropTypes.object,
+  children: PropTypes.node.isRequired,
+  target: PropTypes.any.isRequired,
+  rotated: PropTypes.bool.isRequired,
+  arrowComponent: PropTypes.oneOfType([PropTypes.func, PropTypes.object]).isRequired,
+};
+
+OverlayBase.defaultProps = {
+  className: undefined,
+  classes: undefined,
```
MaximKudriavtsev: Same with `classes` here

PR #2239, opened 2019-08-15
chore(react-grid-demos): switch to useReducer, fix filtering

ushkal:


packages/dx-react-grid-demos/src/demo-sources/grid-lazy-loading/remote-data.jsxt
```diff
@@ -42,7 +44,70 @@ const DateTypeProvider = props => (
...
+        ...state,
+        lastQuery: payload,
+      };
+    default:
+      return state;
+  }
+}
+
 export default () => {
+  const initialState = {
```
MaximKudriavtsev: Should `InitialState` be inside the render method?

ushkal: We can move it outside

PR #2256, opened 2019-08-21
fix(react-grid): correct calculating start index of loading row for Infinite Scrolling

### LINKED ISSUE #2139: Infinite scrolling sometimes won't render rows

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

Sometimes when scrolling fast with Infinite Scrolling enabled the table won't render rows, as shown in the two images below taken from your own documentation.

image

image

There are some other issues with this problem but they have both been closed.
https://github.com/DevExpress/devextreme-reactive/issues/1253
https://github.com/DevExpress/devextreme-reactive/issues/2062

LazyLahtak:
Fixes #2139

### DIFF
--- a/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
@@ -169,7 +169,6 @@ describe('VirtualTableState helpers', () => {
         const loadedInterval = createInterval(200, 500);
         const newInterval = createInterval(100, 400);
         const virtualRows = createVirtualRows(loadedInterval);
-        const referenceIndex = 60;
 
         expect(calculateRequestedRange(virtualRows, newInterval, pageSize))
           .toEqual({ start: 100, end: 200 });
@@ -258,15 +257,33 @@ describe('VirtualTableState helpers', () => {
       });
 
       // tslint:disable-next-line: max-line-length
-      it('should caclulate correct if reference index more than the start of a new interval and less than half of page', () => {
+      describe('reference index more than the start of a new interval and less than half of page', () => {
         const loadedInterval = createInterval(100, 400);
         const newInterval = createInterval(200, 500);
         const virtualRows = createVirtualRows(loadedInterval);
         const referenceIndex = 320;
 
-        expect(calculateRequestedRange(virtualRows, newInterval, pageSize, referenceIndex))
+        it('should caclulate correct if infinite scrolling', () => {
+          expect(calculateRequestedRange(virtualRows, newInterval, pageSize, referenceIndex, true))
           .toEqual({ start: 300, end: 400 });
+        });
+
+        it('should caclulate correct if non-infinite scrolling', () => {
+          expect(calculateRequestedRange(virtualRows, newInterval, pageSize, referenceIndex, false))
+          .toEqual({ start: 400, end: 500 });
+        });
       });
+    });
+
+    // tslint:disable-next-line: max-line-length
+    it('should caclulate correct in non-infinite scrolling', () => {
+      const loadedInterval = createInterval(100, 400);
+      const newInterval = createInterval(200, 500);
+      const virtualRows = createVirtualRows(loadedInterval);
+      const referenceIndex = 320;
+
+      expect(calculateRequestedRange(virtualRows, newInterval, pageSize, referenceIndex))
+        .toEqual({ start: 400, end: 500 });
     });
   });
 
--- a/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
@@ -161,7 +161,7 @@ describe('VirtualTableState helpers', () => {
         const newInterval = createInterval(200, 500);
         const virtualRows = createVirtualRows(loadedInterval);
 
-        expect(calculateRequestedRange(virtualRows, newInterval, 310, pageSize))
+        expect(calculateRequestedRange(virtualRows, newInterval, pageSize))
           .toEqual({ start: 400, end: 500 });
       });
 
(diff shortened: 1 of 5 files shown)


LazyLahtak:
FPR

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
```diff
@@ -43,12 +43,18 @@ export const mergeRows: MergeRowsFn = (
...
-  virtualRows, newRange, pageSize,
+  virtualRows, newRange, pageSize, referenceIndex,
 ) => {
   const loadedInterval = intervalUtil.getRowsInterval(virtualRows);
   const isAdjacentPage = Math.abs(loadedInterval.start - newRange.start) < 2 * pageSize;
   if (isAdjacentPage) {
-    return intervalUtil.difference(newRange, loadedInterval);
+    const calculatedRange = intervalUtil.difference(newRange, loadedInterval);
+    const firstHalfOfPage = calculatedRange.start - referenceIndex > pageSize / 2;
+    if (calculatedRange.start > referenceIndex && firstHalfOfPage) {
```
MaximKudriavtsev: Can we use only the next `calculatedRange.start - referenceIndex > pageSize / 2` codition?

if the above condition is true and `pageSize` > 0 this condition `calculatedRange.start > referenceIndex` is true

LazyLahtak:
FPR

LazyLahtak:
FPR

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
```diff
@@ -143,7 +143,7 @@ export const needFetchMorePages: PureComputed<[VirtualRows, number, number], boo
   const { start, end } = intervalUtil.getRowsInterval(virtualRows);
   const loadCount = end - start;
   const topTriggerIndex = start > 0 ? start + pageSize! : 0;
-  const bottomTriggerIndex = end - pageSize!;
+  const bottomTriggerIndex = Math.max(topTriggerIndex + pageSize, end - pageSize! * 1.5);
```
MaximKudriavtsev: What is a `1.5` value?

LazyLahtak: Half of page.
Sometimes difference between `bottomTriggerIndex` and `topTriggerIndex` more than one `pageSize` and we can get index, than no trigger update. This is index in interval between `end - pageSize!` and `end - pageSize! * 1.5`
For example, in master try after load press `End` button twice (lazy loading demo)
You will stuck at string 150 as last with `bottomTriggerIndex` === 150 and `index ~ 140`

LazyLahtak: Sometimes we may catch this bug not in beginning of table. It is quite difficult to catch, but can.

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
```diff
@@ -228,10 +229,45 @@ describe('VirtualTableState helpers', () => {
...
+
+      // tslint:disable-next-line: max-line-length
+      it('should caclulate correct if reference index more than the start of a new interval and less than half of page', () => {
+        const loadedInterval = createInterval(100, 400);
+        const newInterval = createInterval(200, 500);
+        const virtualRows = createVirtualRows(loadedInterval);
+        const referenceIndex = 320;
+
+        expect(calculateRequestedRange(virtualRows, newInterval, pageSize, referenceIndex))
+          .toEqual({ start: 300, end: 400 });
```
ushkal: Why do we want rows 300-400 if they're loaded already?

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
```diff
@@ -43,12 +43,17 @@ export const mergeRows: MergeRowsFn = (
...
+  virtualRows, newRange, pageSize, referenceIndex,
 ) => {
   const loadedInterval = intervalUtil.getRowsInterval(virtualRows);
   const isAdjacentPage = Math.abs(loadedInterval.start - newRange.start) < 2 * pageSize;
-  if (isAdjacentPage) {
-    return intervalUtil.difference(newRange, loadedInterval);
+  const calculatedRange = intervalUtil.difference(newRange, loadedInterval);
+  if (isAdjacentPage && calculatedRange !== intervalUtil.empty) {
+    if (calculatedRange.start - referenceIndex > pageSize / 2) {
+      calculatedRange.start -= pageSize;
```
ushkal: In non-infinite scrolling mode request range shouldn't be affected

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.test.ts
LazyLahtak: When we try load next page (rows 400-500) width index less than page 350 (for `pageSize === 100`), we catch index bug. It loads nextPage instead curren when we fast scroll up.
We see at string 320, cache know about previous (200-300) and next page (400-500), and we load 400-500. Then, we get rows combination:
```
[200,
201,
...
300,
400,
401,
...
500,
401,
...
500]
```
For index less than `start - pageSize/2` we need to shift loading interval.

ushkal: Does it reproduce in non-infinite mode?

packages/dx-grid-core/src/plugins/virtual-table-state/helpers.ts
LazyLahtak: You're right, I'll fix it.

LazyLahtak:
FPR

LazyLahtak:
FPR

LazyLahtak:
FPR

PR #2257, opened 2019-08-21
fix(react-grid): correctly handle column count changes in virtual table

### LINKED ISSUE #2248: Virtual table in combination with column chooser gives error

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

`VirtualTable` in combination with `ColumnChooser` gives an error when changing column visibility. It works fine when a normal Table is used.

### Expected Behaviour

That `VirtualTable` works the same as normal `Table` when changing column visibility.

### Steps to Reproduce

As mentioned above, when changing column visibility on a `VirtualTable` an error is thrown:
`Uncaught TypeError: Cannot read property 'type' of undefined`
See screenshot below for more details

### Screenshots

[https://imgur.com/a/zlEbc42](https://imgur.com/a/zlEbc42)

### Environment

- devextreme-reactive: 2.0.3
- react: 16.9.0
- browser: Chrome 76.0.3809.100 (64bit)
- bootstrap: none
- react-bootstrap: none
- material-ui: 4.3.2

### LINKED ISSUE #2250: Undefined row id when using remote grouping example

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

I've copy/pasted the "Remote Grouping" example situated [here](https://devexpress.github.io/devextreme-reactive/react/grid/docs/guides/grouping/) in this [codesandbox](https://codesandbox.io/s/devextreme-react-grid-for-material-ui-iyn5l).

After a drag and drop of a column in the grouping header, the message "The row id is undefined. Check the getRowId function. The row is ..." appear in the console. At this point the grouping seems to work fine but if you remove the column from the grouping, theres a "Encountered two children with the same key" which in turn, make the row data undefined.

When you do this on the documentation website, the grid disappear.

I've tried using older version of devexpress in the sandbox but theres no difference.

### Expected Behaviour

Removing a row that was previously grouped from the TableGroupRow should put the table in it's initial state.

### Steps to Reproduce

Drag and drop any column in the group header and then remove it.

### Environment

- devextreme-reactive: 2.0.3
- react: 16.9.0
- browser: chrome
- material-ui: 4.3.2

ushkal:
fixes #2248 
fixes #2250

### DIFF
--- a/packages/dx-grid-core/src/plugins/virtual-table/helpers.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table/helpers.ts
@@ -39,11 +39,17 @@ export const getViewport: GetViewportFn = (
   // NOTE: prevent unnecessary updates
   // e.g. when rows changed but bounds remain the same.
   let result = viewport;
-  if (viewportTop !== viewport.viewportTop) {
-    result = { ...result, viewportTop };
+  if (viewportTop !== viewport.top) {
+    result = { ...result, top: viewportTop };
   }
-  if (viewportLeft !== viewport.viewportLeft) {
-    result = { ...result, viewportLeft };
+  if (viewportLeft !== viewport.left) {
+    result = { ...result, left: viewportLeft };
+  }
+  if (containerWidth !== viewport.width) {
+    result = { ...result, width: containerWidth };
+  }
+  if (containerHeight !== viewport.height) {
+    result = { ...result, height: containerHeight };
   }
   if (!arraysEqual(rows, viewport.rows)) {
     result = { ...result, rows };
--- a/packages/dx-grid-core/src/plugins/virtual-table/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table/helpers.test.ts
@@ -35,8 +35,10 @@ describe('#getViewport', () => {
       rows: [0, 0],
       headerRows: [0, 0],
       footerRows: [0, 0],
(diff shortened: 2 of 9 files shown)


packages/dx-grid-core/src/plugins/virtual-table/helpers.ts
```diff
@@ -39,11 +39,17 @@ export const getViewport: GetViewportFn = (
...
-  if (viewportLeft !== viewport.viewportLeft) {
-    result = { ...result, viewportLeft };
+  if (viewportLeft !== viewport.left) {
+    result = { ...result, left: viewportLeft };
+  }
+  if (containerWidth !== viewport.width) {
+    result = { ...result, width: containerWidth };
+  }
+  if (containerHeight !== viewport.height) {
+    result = { ...result, height: containerHeight };
```
MaximKudriavtsev: Can we use the `return` on each `if` condition?

ushkal: No, we assemble viewport by pieces

MaximKudriavtsev: This is a horror 😱

ushkal: The way to avoid rerenders

PR #2265, opened 2019-08-22
fix(react-grid): allow DataTypeProvider editor to know if editing is enabled

### LINKED ISSUE #2247: Filter editor can't be disabled when DataTypeProvider is used

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

I'm using a table with both `TableEditRow` and `TableFilterRow` as well as a `DataTypeProvider` to add a boolean data type (to render a Yes/No dropdown selection as the `editorComponent`). Problem is, I only wanted this to render in TableEditRow and not TableFilterRow, but it's rendering it for both. I'm also not receiving the `disabled` prop in my `editorComponent` even though I have `filteringDisabled` in `columnExtensions`. 

### Expected Behaviour

Properly receive column extension props in `editorComponent` of `DataTypeProvider`

ushkal:
fixes #2247 

Extends [ValueEditorProps ](https://devexpress.github.io/devextreme-reactive/react/grid/docs/reference/data-type-provider/#datatypeprovidervalueeditorprops)with the `disabled `prop


packages/dx-react-grid/src/types/data-type-provider/data-type-provider.types.ts
```diff
@@ -22,6 +22,8 @@ export namespace DataTypeProvider {
     value: any;
     /** Handles value changes. */
     onValueChange: (newValue: any) => void;
+    /** Specifies whether editing a column is enabled. */
+    editingEnabled: boolean;
```
MaximKudriavtsev: Should we make this property public?

ushkal: It is public :)

MaximKudriavtsev: I mean - should we add this property into our documentation?

ushkal: Yes, I'll add it

packages/dx-react-grid/docs/reference/data-type-provider.md
```diff
@@ -47,6 +47,7 @@ column | [Column](grid.md#column) | A column object.
 row? | any | A row.
 value | any | Specifies the editor value.
 onValueChange | (newValue: any) => void | Handles value changes.
+editingEnabled | boolean | Specifies whether editing a column is enabled.
```
MaximKudriavtsev: Maybe rename this property to `readOnly`? Because this name says about editing only.

ushkal:
FPR

packages/dx-react-grid/docs/reference/data-type-provider.md
```diff
@@ -47,6 +47,7 @@ column | [Column](grid.md#column) | A column object.
 row? | any | A row.
 value | any | Specifies the editor value.
 onValueChange | (newValue: any) => void | Handles value changes.
+disabled | boolean | Specifies whether the editor is disabled.
```
RomanTsukanov:
```suggestion
disabled | boolean | **true** if users should not be able to edit the value, **false** otherwise.
```

packages/dx-react-grid/src/types/data-type-provider/data-type-provider.types.ts
```diff
@@ -22,6 +22,8 @@ export namespace DataTypeProvider {
     value: any;
     /** Handles value changes. */
     onValueChange: (newValue: any) => void;
+    /** Specifies whether the editor is enabled. */
```
RomanTsukanov:
```suggestion
    /** "true" if users should not be able to edit the value, "false" otherwise. */
```

PR #2267, opened 2019-08-22
fix(react-chart): fix demo with export in production mode

Krijovnick:


packages/dx-react-common/package.json
```diff
@@ -45,7 +45,7 @@
     "nodemon": "^1.19.1",
     "source-map-loader": "^0.2.4",
     "style-loader": "^1.0.0",
-    "webpack": "^4.39.2",
+    "webpack": "^4.38.0",
```
MaximKudriavtsev: If you are want to fix this version, you should remove the `^` symbol from all. And remove `yarn.lock` file and fire `yarn` command.

PR #2272, opened 2019-08-23
fix(react-scheduler): fix create an appointment on the first call of AppointmentForm

### LINKED ISSUE #2271: Appointment cannot be created in the first form though it can be created in the following forms

### I'm using ...

- [x] React Scheduler

### Current Behaviour
Appointment cannot be created in the first form that was called in Editing mode in Scheduler.

### Expected Behaviour
Appointment should be created

### Steps to Reproduce
1) Call AppointmentForm.
2) Save a new Appointment.
3) Appointment hasn't been saved.

### Screenshots
Appointment is not created

### Environment

- devextreme-reactive: 2.0.3

AryamnovEugeniy:
Fixes #2271

### DIFF
--- a/packages/dx-react-scheduler/src/plugins/editing-state.tsx
+++ b/packages/dx-react-scheduler/src/plugins/editing-state.tsx
@@ -28,7 +28,7 @@ class EditingStateBase extends React.PureComponent<EditingStateProps, EditingSta
   commitDeletedAppointment: ActionFn<any>;
 
   static defaultProps: Partial<EditingStateProps> = {
-    defaultEditingAppointment: {},
+    defaultEditingAppointment: undefined,
     defaultAppointmentChanges: {},
     defaultAddedAppointment: {},
     preCommitChanges: preCommitChangesBase,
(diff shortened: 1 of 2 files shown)


packages/dx-react-scheduler/src/plugins/editing-state.tsx
```diff
@@ -28,7 +28,7 @@ class EditingStateBase extends React.PureComponent<EditingStateProps, EditingSta
   commitDeletedAppointment: ActionFn<any>;
 
   static defaultProps: Partial<EditingStateProps> = {
-    defaultEditingAppointment: {},
+    defaultEditingAppointment: undefined,
     defaultAppointmentChanges: {},
     defaultAddedAppointment: {},
```
MaximKudriavtsev: Should we apply this change to other fields?

AryamnovEugeniy: No< we shouldn't. These should be empty objects in order to be able to edit correctly.

PR #2290, opened 2019-08-30
fix(scheduler-core): display recurrent appointments in the lower right corner of WeekView correctly

### LINKED ISSUE #2287: Recurrent appointment disappears in the lower right corner of WeekView.

### I'm using ...

- [x] React Scheduler

### Current Behaviour
Recurrent appointment disappears in the lower left corner of WeekView.

### Expected Behaviour
It shouldn't disappear.

### Steps to Reproduce
Place recurrent appointment so it would be in the lower right corner of WeekView.

### Screenshots
Second bug

### Environment

- devextreme-reactive: 2.0.3

AryamnovEugeniy:
Fixes #2287

### DIFF
--- a/packages/dx-scheduler-core/src/utils.test.ts
+++ b/packages/dx-scheduler-core/src/utils.test.ts
@@ -610,19 +610,15 @@ describe('Utils', () => {
       expect(result[1].end.toString())
         .toBe(moment(new Date('2019-04-26T13:00:00+0600')).toString());
     });
-    fit('should work correctly with near-boundary appointments', () => {
+    it('should work correctly with near-left-boundary appointments', () => {
       const leftBoundary = new Date('2019-04-25 00:00:00+0300');
       const rightBoundary = new Date('2019-04-26 23:59:00+0300');
-      const appointments = [{
+      const appointment = {
         start: moment(new Date('2019-04-25T00:00:00+0300')),
         end: moment(new Date('2019-04-25T13:00:00+0300')),
         rRule: 'FREQ=DAILY;COUNT=2',
-      }, {
-        start: moment(new Date('2019-04-25T12:11:00+0300')),
-        end: moment(new Date('2019-04-25T23:00:00+0300')),
-        rRule: 'FREQ=DAILY;COUNT=2',
-      }];
-      let result = filterByViewBoundaries(appointments[0], leftBoundary, rightBoundary);
+      };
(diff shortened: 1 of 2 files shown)


packages/dx-scheduler-core/src/utils.test.ts
```diff
@@ -610,5 +610,45 @@ describe('Utils', () => {
       expect(result[1].end.toString())
         .toBe(moment(new Date('2019-04-26T13:00:00+0600')).toString());
     });
+    fit('should work correctly with near-boundary appointments', () => {
```
MaximKudriavtsev: fit -> it

packages/dx-scheduler-core/src/utils.test.ts
```diff
@@ -610,5 +610,45 @@ describe('Utils', () => {
...
+        .toBe(moment(new Date('2019-04-25T00:00:00+0300')).toString());
+      expect(result[0].end.toString())
+        .toBe(moment(new Date('2019-04-25T13:00:00+0300')).toString());
+
+      expect(result[1].start.toString())
+        .toBe(moment(new Date('2019-04-26T00:00:00+0300')).toString());
+      expect(result[1].end.toString())
+        .toBe(moment(new Date('2019-04-26T13:00:00+0300')).toString());
+
+      result = filterByViewBoundaries(appointments[1], leftBoundary, rightBoundary);
```
MaximKudriavtsev: Can we reduce it into two tests? I think it would look simply

packages/dx-scheduler-core/src/utils.ts
```diff
@@ -307,6 +307,8 @@ export const calculateRectByDateIntervals: CalculateRectByDateIntervalsFn = (
 const expandRecurrenceAppointment = (
   appointment: AppointmentMoment, leftBound: Date, rightBound: Date,
 ) => {
+  const rightBoundUTC = new Date(getUTCDate(rightBound));
+  const leftBoundUTC = new Date(getUTCDate(leftBound) - 1);
```
MaximKudriavtsev: Why we are using `-1` here?

MaximKudriavtsev: Can we use `inc=true` instead of `-1`? See doc [here](https://github.com/jakubroztocil/rrule#rruleprototypebetweenafter-before-incfalse--iterator).

PR #2291, opened 2019-08-30
fix(react-grid): allow columns bands to work in virtual table

### LINKED ISSUE #1389: Banded columns and virtual table bug

ushkal:
fixes #1389


packages/dx-grid-core/src/plugins/table-band-header/computeds.ts
```diff
@@ -73,3 +75,67 @@ export const tableHeaderColumnChainsWithBands: GetHeaderColumnChainsFn<
...
+const getBandLevel: PureComputed<[any[], string, number?], number> = (
+  bands, bandTitle, level = 0,
+) => {
+  for (const band of bands) {
+    if (band.title === bandTitle) {
+      return level;
+    }
+    if (band.children !== undefined) {
+      const result = getBandLevel(band.children, bandTitle, level + 1);
+      if (result >= 0) return result;
```
MaximKudriavtsev: We can include `result >= 0` into above `if`

ushkal: I can't see how. Can you please expand your idea?

MaximKudriavtsev: My bed, you are right

packages/dx-grid-core/src/plugins/table-band-header/computeds.ts
```diff
@@ -73,3 +75,67 @@ export const tableHeaderColumnChainsWithBands: GetHeaderColumnChainsFn<
 
   return [...bandChains, ...chains.slice(maxBandRowIndex)];
 };
+
+const getBandLevel: PureComputed<[any[], string, number?], number> = (
+  bands, bandTitle, level = 0,
```
MaximKudriavtsev: Should we use the type `ColumnBands` instead of `any[]`.

packages/dx-grid-core/src/plugins/table-band-header/computeds.test.ts
```diff
@@ -140,4 +143,103 @@ describe('TableBandHeader Plugin computeds', () => {
       );
     });
   });
+
+  describe('#columnBandLevels', () => {
+    it('should in simple case', () => {
```
MaximKudriavtsev: "Should work..." ?

PR #2332, opened 2019-09-16
fix(react-grid): correct displaying "No data" message in Edge

### LINKED ISSUE #2330: "No data" message is incorrectly displayed in Edge
When we use Fixed Columns and have no rows to display, the "No data" message is displayed in the wrong place.

image

LazyLahtak:
Fixes #2330


packages/dx-react-grid-bootstrap3/src/templates/table-no-data-cell.jsx
```diff
@@ -1,20 +1,7 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import { getStickyPosition } from '../utils/css-fallback-properties';
 
 export class TableNoDataCell extends React.PureComponent {
```
MaximKudriavtsev: Should we keep the `class` here if we don't use the `constructor`?

packages/dx-react-grid-bootstrap3/src/templates/table-no-data-cell.jsx
```diff
@@ -38,7 +24,7 @@ export class TableNoDataCell extends React.PureComponent {
         <div
           style={{
             display: 'inline-block',
-            position: stickyPosition,
+            position: 'fixed',
```
MaximKudriavtsev: We can extract it into constant

LazyLahtak: What does you mean? It is normal CSS value.

MaximKudriavtsev:
```js
const style = {
  display: 'inline-block',
  position: 'fixed',
  // and so on...
};

// ...

<div
  style={style}
>
```

LazyLahtak: Should we do that? We use that style once, in this `div`

MaximKudriavtsev: We create a new object on each render. We often try to avoid similar cases.

packages/dx-react-grid-bootstrap3/src/templates/table-no-data-cell.jsx
```diff
@@ -1,62 +1,42 @@
...
+      padding: '50px 0',
+      ...style,
+    }}
+    colSpan={colSpan}
+    {...restProps}
+  >
+    <div
+      style={{
+        display: 'inline-block',
+        position: 'fixed',
```
ushkal: 'fixed' position is a bad idea because it's relative to a page not a grid

LazyLahtak: When "No data" message is relative to grid, we get wrong result (see GIF). If we should'n make some relate to page, we need to relate on grid's container.
Also is message relative to page critical? I tried to resize, zoom and move window and didn't break this.
GIF123123123123123123123

ushkal: Well, try to set grid's width to 300px and a page height to 3000px

LazyLahtak: I get it, I'll try another.

LazyLahtak:
FPR

LazyLahtak:
FPR

PR #2336, opened 2019-09-17
refactor(scheduler-core): refactor EditingState helpers

AryamnovEugeniy:


packages/dx-scheduler-core/src/plugins/editing-state/helpers.test.ts
```diff
@@ -71,6 +71,24 @@ describe('EditingState', () => {
         const changes = deleteCurrent(appointmentData);
         expect(changes).toEqual({ deleted: 0 });
       });
+
+      it('should remove all sequence if current item is last', () => {
```
MaximKudriavtsev: This test is duplicated with previous.

PR #2364, opened 2019-09-27
fix(react-grid-bootstrap4): correct position when horizontal scrolling

LazyLahtak:


packages/dx-styles/dx-grid-bs4.scss
```diff
@@ -218,7 +218,7 @@ button.dx-g-bs4-table-edit-command-cell {
   z-index: 300;
   background-clip: padding-box;
 }
-td.dx-g-bs4-fixed-cell{
+.dx-g-bs4-no-data-fixed-cell{
```
ushkal: dx-g-bs4-no-data-fixed-cell > dx-g-bs4-no-data-cell

PR #2372, opened 2019-09-30
feat(scheduler-core): use different default counts for different recurrence types

AryamnovEugeniy:


packages/dx-scheduler-core/src/plugins/appointment-form/utils.ts
```diff
@@ -67,3 +68,22 @@ export const getAvailableRecurrenceOptions: OptionsFormatterFn = getMessage =>
...
+    case RRULE_REPEAT_TYPES.DAILY:
+      return BASIC_DAILY_COUNT;
+    default:
+      return 1;
+  }
+};
+
+export const checkNumber: PureComputed<
+  [number], boolean
+> = number => number > 0 && number <= Number.MAX_SAFE_INTEGER;
```
MaximKudriavtsev: This function has called `checkNumber` but negative values are also numbers. I think we should change the name

packages/dx-scheduler-core/src/plugins/appointment-form/utils.test.ts
```diff
@@ -151,4 +152,36 @@ describe('AppointmentForm utils', () => {
...
     });
   });
+  describe('#getCountDependingOnRecurrenceType', () => {
+    it('should return count = 30 for daily appointments', () => {
+      expect(getCountDependingOnRecurrenceType(RRULE_REPEAT_TYPES.DAILY))
+        .toEqual(BASIC_DAILY_COUNT);
+    });
+    it('should return count = 13 for weekly appointments', () => {
+      expect(getCountDependingOnRecurrenceType(RRULE_REPEAT_TYPES.WEEKLY))
+        .toEqual(BASIC_WEEKLY_COUNT);
```
MaximKudriavtsev: We should expect `13` instead of `BASIC_WEEKLY_COUNT`. Because if constant will change, you can't know about it.

packages/dx-scheduler-core/src/plugins/appointment-form/helpers.test.ts
```diff
@@ -141,6 +150,7 @@ describe('AppointmentForm helpers', () => {
           freq: RRule.DAILY,
           bymonthday: undefined,
           byweekday: undefined,
+          count: getCountDependingOnRecurrenceType(RRule.DAILY),
```
MaximKudriavtsev: !

AryamnovEugeniy:
FPR

PR #2390, opened 2019-10-03
fix(react-grid): allow inline cell editing only for table data type rows

### LINKED ISSUE #2389: Incorrect behavior of the TableEditRow with the TableInlineCellEditing plugin
When the TableInlineCellEditing plugin is placed after the TableEditRow plugin, cells in added rows lose focus. It makes impossible to type any character.

LazyLahtak:
Fixes #2389


LazyLahtak:
FPR

PR #2392, opened 2019-10-03
fix(react-grid-material-ui): add padding for first edit cell in in-line cell editing mode

### LINKED ISSUE #2391: Incorrect size of the first cell in in-line cell editing mode
In Material UI, the first cell in a row changes its width if it is switched to the edit mode using the TableInlineCellEditing plugin.

image

LazyLahtak:
Fixes #2391


LazyLahtak:
FPR

PR #2399, opened 2019-10-04
fix(react-grid): check next column name by column.name instead type

### LINKED ISSUE #2397: Wrong render of cell`s separators for resizing
When the TableColumnResizing has `nextColumn` resizingMode, the grid may render a separator to resize the column which isn't allowed to be resized. An attempt to resize that column crashes the project.

LazyLahtak:
Fixes #2397

### DIFF
--- a/packages/dx-react-grid/src/plugins/table-header-row.tsx
+++ b/packages/dx-react-grid/src/plugins/table-header-row.tsx
@@ -8,7 +8,6 @@ import {
   tableRowsWithHeading,
   isHeadingTableCell,
   isHeadingTableRow,
-  getLastColumnName,
   getNextColumnName,
   TABLE_DATA_TYPE,
   TABLE_HEADING_TYPE,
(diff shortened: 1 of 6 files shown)


LazyLahtak:
FPR

packages/dx-react-grid/src/plugins/table-header-row.tsx
```diff
@@ -8,7 +8,6 @@ import {
   tableRowsWithHeading,
   isHeadingTableCell,
   isHeadingTableRow,
-  getLastColumnName,
```
MaximKudriavtsev: We should also remove this function from tests. (mock and import)

packages/dx-react-grid/src/plugins/table-header-row.tsx
```diff
@@ -94,7 +92,7 @@ class TableHeaderRowBase extends React.PureComponent<TableHeaderRowProps> {
                     draggingEnabled={draggingEnabled && atLeastOneDataColumn}
                     resizingEnabled={
                       tableColumnResizingEnabled
-                      && (!lastColumn || columnResizingMode === 'widget')
+                      && (!!nextColumnName || columnResizingMode === 'widget')
```
MaximKudriavtsev: Note: very wired using `widget` as a sting in a lot of places instead of using one constant.

LazyLahtak:
FPR

PR #2439, opened 2019-10-16
fix(react-scheduler): make it impossible to edit radio groups in read-only mode and set default interval to 1

### LINKED ISSUE #2438: It is possible to edit radio groups in AppointmentForm in read-only mode
### I'm using ...

- [ ] React Grid
- [ ] React Chart
- [x] React Scheduler

### Current Behaviour
image

### Expected Behaviour
image

### Steps to Reproduce
```
<AppointmentForm readOnly />
```

AryamnovEugeniy:
Fixes #2438


packages/dx-react-scheduler-material-ui/src/templates/appointment-form/recurrence/layouts/daily.jsx
```diff
@@ -36,7 +37,7 @@ export const Daily = ({
       textEditorComponent={textEditorComponent}
       labelComponent={labelComponent}
       changeRecurrenceInterval={changeRecurrenceInterval}
-      interval={recurrenceOptions.interval}
+      interval={getRecurrenceInterval(recurrenceOptions.interval)}
```
MaximKudriavtsev: Why are you validate `interval` here instead of validating it in `defaultProps`?

AryamnovEugeniy:
FPR

PR #2464, opened 2019-10-17
fix(react-grid): handle column count decrease in virtual table

### LINKED ISSUE #2447: Grid crashes when column count is changed if the TableFilterRow plugin is used

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour
TableFilterRow plugin crashes Grid. 

### Expected Behaviour
Grid should work as without it.

### Steps to Reproduce
1. Open sample https://codesandbox.io/s/lucid-fermat-50vcb
2. Scroll the grid to the very right.
3. Press the  " Set DATA_SET_2" button under the grid.

Application crashes.
Сrash is reproduced if TableFilterRow  plugin is connected.

Sample 1 that crashes : https://codesandbox.io/s/lucid-fermat-50vcb
Sample 2 without TableFilterRow that doesn't crash https://codesandbox.io/s/restless-microservice-08tb4

### Environment

- devextreme-reactive: 2.0.2
- react: 16.8.6
- browser: Google Chrome
- bootstrap: none
- react-bootstrap: none
- material-ui: none

LazyLahtak:
Fixes #2447


packages/dx-grid-core/src/utils/virtual-table.ts
```diff
@@ -113,6 +113,9 @@ export const getSpanBoundary: GetSpanBoundaryFn = (
     let [start] = visibleBoundary;
     const endIndex = Math.min(visibleBoundary[1], items.length - 1);
     let end = endIndex;
+    if (start > end) {
+      start = 0;
+    }
```
MaximKudriavtsev:
```js
const endIndex = Math.min(visibleBoundary[1], items.length - 1);
let end = endIndex;
let start = visibleBoundary < end ? visibleBoundary[0] : 0;
```

PR #2484, opened 2019-10-23
fix(react-grid-material-ui): correct default properties for Virtual Table

### LINKED ISSUE #2482: Virtual Table incorrectly loads rows on scroll up
When we scroll up using the mouse wheel, rows are loaded incorrectly and we see a skeleton cell.

VTBug

LazyLahtak:
Fixes #2482


LazyLahtak:
FPR

PR #2531, opened 2019-11-12
refactor(react-chart): remove redundant files for tooltip

Krijovnick:


packages/dx-react-chart-demos/src/demo-sources/bar-chart/material-ui/side-by-side.jsx
```diff
@@ -99,8 +77,6 @@ export default class Demo extends React.PureComponent {
           <Legend position="bottom" rootComponent={Root} labelComponent={Label} />
           <Title text="Olimpic Medals in 2008" />
           <Stack />
-          <EventTracker />
```
MaximKudriavtsev: Why have we removed `EvetnTracker`?

Krijovnick: In its original form there is no `EventTracker` and `Tooltip`. They have been added accidentally.

PR #2566, opened 2019-11-29
chore: support React DevTools for demo page

MoonBall:
Because a demo is in a iframe element, we can't select a component by React DevTools.
This change refer to https://github.com/facebook/react-devtools/issues/76.


MaximKudriavtsev:
@MoonBall 

Thank you!

PR #2596, opened 2019-12-12
fix(react-scheduler-material-ui): do not show comma in zero-time appointments' title if the title is undefined

### LINKED ISSUE #2595: Unnecessary comma in zero-time appointments without title
When we create a new appointment zero-time appointment without title, there is unnecessary comma.

image

AryamnovEugeniy:
Fixes #2595


packages/dx-react-scheduler-material-ui/src/templates/utils.js
```diff
@@ -42,3 +42,5 @@ export const getAppointmentColor = (level, color, defaultColor) => {
   if (typeof color === 'string') return color;
   return ensureColor(level, color);
 };
+
+export const addCommaAndSpaceToString = string => !!string && `${string},\xa0`;
```
MaximKudriavtsev: Should we return a non-bulean value? 
image

```js
export const addCommaAndSpaceToString = string => string ? `${string},\xa0` : null;
```

PR #2597, opened 2019-12-12
fix(react-grid): Adjust readonly property to match Typescript spec

justincasey:
Fix to the usage of `readonly` attribute on interfaces/types.


ushkal:
Hi @justincasey ,

Thank you for the contribution!

justincasey:
Thanks @ushkal

PR #2602, opened 2019-12-17
refactor(react-scheduler): use getter instead of template to pass parameters to appointments

AryamnovEugeniy:


packages/dx-scheduler-core/src/utils.ts
```diff
@@ -320,14 +323,17 @@ export const calculateRectByDateIntervals: CalculateRectByDateIntervalsFn = (
   const { growDirection, multiline } = type;
   const isHorizontal = growDirection === HORIZONTAL_TYPE;
 
-  const sorted = sortAppointments(intervals);
-  const grouped = findOverlappedAppointments(sorted as AppointmentMoment[], isHorizontal);
+  const sorted = intervals.map(intervalGroup => sortAppointments(intervalGroup));
```
MaximKudriavtsev:
```js
const sorted = intervals.map(intervalGroup => sortAppointments(intervalGroup));
// same with
const sorted = intervals.map(sortAppointments);
```

packages/dx-react-scheduler/src/plugins/month-view.tsx
```diff
@@ -1,26 +1,20 @@
 import * as React from 'react';
 import { Plugin, PluginComponents } from '@devexpress/dx-react-core';
-import { monthCellsData, horizontalTimeTableRects } from '@devexpress/dx-scheduler-core';
+import { monthCellsData, calculateMonthDateIntervals as calculateMonthDateIntervalsCore } from '@devexpress/dx-scheduler-core';
```
ushkal: We don't need to rename it to ...Core

AryamnovEugeniy:
FPR

packages/dx-react-scheduler/src/plugins/resources.tsx
```diff
@@ -1,14 +1,25 @@
...
 import { ResourcesProps } from '../types/resources/resources.types';
-import { Appointments } from '../types';
+import { memoize } from '@devexpress/dx-core';
 
 const pluginDependencies = [
   { name: 'Appointments' },
 ];
 
-const ResourcesBase: React.SFC<ResourcesProps> = ({
+const addResourcesToTimeTableAppointments = memoize(({
```
MaximKudriavtsev: We should not use memoization in the global context

Issue #2607, opened 2019-12-18
Cannot read property 'start' of undefined


SMaverick78:
Hello! I have tables in tabs. When switching tabs I get an error. ((((
breadcrumbs.js:64 TypeError: Cannot read property 'start' of undefined
    at ee (dx-grid-core.es.js:1882)
    at dx-react-grid.es.js:2707
    at e.render (dx-react-core.es.js:535)
https://github.com/DevExpress/devextreme-reactive/blob/d83cdc32baef5517b3dcc5c00877cb1f7080f646/packages/dx-grid-core/src/plugins/table-fixed-columns/helpers.ts#L45

ushkal:
Hi @SMaverick78 ,

I couldn't reproduce the described behavior. Please send us a runnable project or modify this [codesandbox](https://codesandbox.io/s/devextreme-react-grid-for-material-ui-0wj8n) so that the issue is reproduced.

ARL64:
Hi @ushkal, I have the same issue. Using VirtualTable, TableRowDetail, RowDetailState and rightColums.
Using Table instead of VirtualTable and leftColumns instead of rightColumns, I get the error.

Unable to reproduce it on codepen because of _A cross-origin error was thrown. React doesn't have access to the actual error object in development._

Here is the [codesanbox](https://codesandbox.io/s/cannot-read-property-start-of-undefined-0ze6l).

Removing the TableRowDetail get it working.

ARL64:
![Capture d’écran 2019-12-19 à 15 33 48](https://user-images.githubusercontent.com/837109/71181764-1231e800-2275-11ea-913c-2618b17ceaee.png)

Index
`import React, { useState } from "react";
import ReactDOM from "react-dom";
import { Tabs, Tab, Typography } from "@material-ui/core";
import { RowDetailState } from "@devexpress/dx-react-grid";
import {
  Grid,
  TableHeaderRow,
  TableFixedColumns,
  TableRowDetail,
  VirtualTable,
} from "@devexpress/dx-react-grid-material-ui";

const TabPanel = props => {
  const { children, value, index, ...other } = props;

  return (
    <Typography
      {...other}
      component="div"
      role="tabpanel"
      hidden={value !== index}
      style={{ height: "100%" }}
      id={`simple-tabpanel-${index}`}
    >
      <div style={{ height: "100%" }}>{children}</div>
    </Typography>
  );
};

function App() {
  const [columns] = useState([
[...]

PR #2614, opened 2019-12-20
fix(react-grid): specify detail row as a dependency for fixed columns

ushkal:
related to https://github.com/DevExpress/devextreme-reactive/issues/2607#issuecomment-567843925


MaximKudriavtsev:
Documentation also?

PR #2616, opened 2019-12-24
chore: add ES2020 features using babel's plugins

LazyLahtak:
Add Nullish Coalescing Operator:
```js
const result = value ?? defaultValue
```
equal
```js
let result;
if (value !== null && value !== undefined) {
  result = value;
} else {
  result = defaultValue;
}
```
_________________________
Add Optional Chaining:
```js
const obj = {
  field: {
    value: 5,
  },
}

obj?.field?.value     // 5
obj?.subfield?.value  //undefined
```

```js
const obj = {
  func: () => { return 'someText'; },
}

obj?.func?.()         // 'someText'
obj?.anotherFunc?.()  // undefined
```

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -9,7 +9,6 @@ module.exports = {
     }],
     "@babel/plugin-proposal-nullish-coalescing-operator",
     "@babel/plugin-proposal-optional-chaining",
-
   ],
   "env": {
     "test": {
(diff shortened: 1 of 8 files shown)


babel.config.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
     ["transform-react-remove-prop-types", {
       "mode": "unsafe-wrap"
     }],
+    "@babel/plugin-proposal-nullish-coalescing-operator",
+    "@babel/plugin-proposal-optional-chaining",
+
```
ushkal: unneeded empty string

MaximKudriavtsev:
CC: @AryamnovEugeniy, @Krijovnick

PR #2618, opened 2019-12-27
fix(react-grid): split band columns with the same name but different parents

### LINKED ISSUE #2617: Band column cant divided correctly in some situation
image

Hi, i got a problem in this picture, you can see the middle line cell melt to one, but they have different father. They are expected to be two cell ! How could i accomplish this feature. any advice would be helpful !

ushkal:
fixes #2617

### DIFF
--- a/packages/dx-grid-core/src/plugins/table-band-header/helpers.ts
+++ b/packages/dx-grid-core/src/plugins/table-band-header/helpers.ts
@@ -16,24 +16,27 @@ export const isBandedOrHeaderRow: IsSpecificRowFn = tableRow => isBandedTableRow
 export const isNoDataColumn = (columnType: symbol) => columnType !== TABLE_DATA_TYPE;
 
 export const getColumnMeta: GetColumnBandMetaFn = (
-  columnName, bands, tableRowLevel,
+  columnName, bands, tableRowLevel, key = '',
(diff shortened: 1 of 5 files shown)


ushkal:
FPR

packages/dx-grid-core/src/plugins/table-band-header/helpers.ts
```diff
@@ -16,24 +16,27 @@ export const isBandedOrHeaderRow: IsSpecificRowFn = tableRow => isBandedTableRow
...
+  columnName, bands, tableRowLevel, key = '',
   level = 0, title = null, result = null,
 ) => bands.reduce((acc, column) => {
   if (column.columnName === columnName) {
-    return { ...acc, title, level };
+    return { ...acc, title, level, key };
   }
   if (column.children !== undefined) {
+    const bandTitle = level > tableRowLevel ? title : column.title;
+    const bandKey = level > tableRowLevel ? key : `${key}_${bandTitle}`;
```
MaximKudriavtsev: Should we extract `level > tableRowLevel` into one more variable?

ushkal: Yep, good point

PR #2635, opened 2020-01-15
chore: fix heap out of memory issue

AryamnovEugeniy:


MaximKudriavtsev:
Should we update our local node version too?

AryamnovEugeniy:
> Should we update our local node version too?

We may update the local version but this is not urgent since we do not suffer from the issue with heap size locally.

PR #2636, opened 2020-01-15
chore: correct scroll-to-row demo

LazyLahtak:
Allow sorting only by the `name` column.
Get rid of sorting lables for other column (using `SortingState.columnExtensions`)

### DIFF
--- a/packages/dx-react-grid-demos/src/demo-sources/grid-virtual-scrolling/scroll-to-row.jsxt
+++ b/packages/dx-react-grid-demos/src/demo-sources/grid-virtual-scrolling/scroll-to-row.jsxt
@@ -19,6 +19,10 @@ import {
 
 const getRowId = row => row.id;
 
+const defaultSorting = [
+  { columnName: 'name', direction: 'asc' },
+];
+
 export default () => {
   const [columns] = useState([
     { name: 'name', title: 'Name' },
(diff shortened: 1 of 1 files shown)


packages/dx-react-grid-demos/src/demo-sources/grid-virtual-scrolling/scroll-to-row.jsxt
```diff
@@ -34,6 +34,14 @@ export default () => {
     { columnName: 'id', width: 75 },
   ]);
   const vtRef = useRef();
+  const [defaultSorting] = useState([
+    { columnName: 'name', direction: 'asc' },
+  ]);
```
MaximKudriavtsev: I think we should define `defaultXXX` properties as a constant instead of `react state` because these properties are not reactive and these changes will not rerender an element.

LazyLahtak: The same for `sortingStateColumnExtensions`?

MaximKudriavtsev: If it is a `reactive` property

PR #2638, opened 2020-01-16
fix(react-scheduler): fix error caused by view switching when all-day panel is used

### LINKED ISSUE #2637: Scheduler - view switching may not work with all-day panel

### I'm using ...

- [ ] React Grid
- [ ] React Chart
- [x] React Scheduler

### Steps to Reproduce
1) Create a Scheduler with the all-day panel and the ViewSwitcher;
2) Switch from one view to another and then back again;
3) This may cause an error.

### Screenshots
switching

AryamnovEugeniy:
Fixes #2637

### DIFF
--- a/packages/dx-scheduler-core/src/plugins/appointments/helpers.test.ts
+++ b/packages/dx-scheduler-core/src/plugins/appointments/helpers.test.ts
@@ -1,0 +1,47 @@
+import { isTimeTableElementsMetaActual, isAllDayElementsMetaActual } from './helpers';
+
+describe('Appointments helpers', () => {
+  describe('#isTimeTableElementsMetaActual', () => {
+    it('should return true if getCellRects exists', () => {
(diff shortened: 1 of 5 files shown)


packages/dx-scheduler-core/src/plugins/appointments/helpers.test.ts
```diff
@@ -0,0 +1,47 @@
...
+      expect(isAllDayElementsMetaActual(viewCellsData, allDayElementsMeta))
+        .toBeTruthy();
+    });
+
+    it('should return false if getCellRects does not exist', () => {
+      expect(isAllDayElementsMetaActual(viewCellsData, {}))
+        .toBeFalsy();
+    });
+
+    it('should return true if getCellRects length is not equal to the first row', () => {
```
ushkal: “Should return false”

PR #2650, opened 2020-01-28
fix(react-grid): allow to change "TableTreeColumn.for" property in runtime

### LINKED ISSUE #2649: TableTreeColumn does not work when changing the FOR property.
https://codesandbox.io/s/devextreme-react-grid-for-material-ui-nef1j

**TableTreeColumn** does not work when changing the **for** property.

### I'm using ...

- [X] React Grid
- [ ] React Chart
- [ ] React Scheduler

LazyLahtak:
Fixes #2649

### DIFF
--- a/packages/dx-react-grid/src/plugins/table-tree-column.test.tsx
+++ b/packages/dx-react-grid/src/plugins/table-tree-column.test.tsx
@@ -226,46 +226,48 @@ describe('TableTreeColumn', () => {
   });
 
   it('should change "for" property in runtime', () => {
-      class Test extends React.Component {
-        constructor(props) {
-          super(props);
-          this.state = {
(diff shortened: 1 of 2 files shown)


packages/dx-react-grid/src/plugins/table-tree-column.test.tsx
```diff
@@ -24,7 +24,7 @@ const defaultDeps = {
     tableCell: {
       tableRow: { type: 'undefined', rowId: 1, row: { value: 'value' } },
       tableColumn: { type: 'undefined', rowId: 1, column: { name: 'a' } },
-      style: {},
+      style: {},
```
ushkal: unnecessary whitespace

PR #2685, opened 2020-02-19
docs(react-grid): data validation guide

ushkal:


LazyLahtak:
Validation On Typing demo
Should we highlight invalid cells (using border or background color)? It seems doesn't clear why `SAVE` is blocked.
image

MaximKudriavtsev:
Should we make a new guide instead of a section? I think it will be better to find information and reduces guide sizes

packages/dx-react-grid-demos/src/demo-sources/grid-editing/validation-on-commit.jsxt
```diff
@@ -0,0 +1,114 @@
...
+
+const columns = [
+  { name: 'firstName', title: 'First Name' },
+  { name: 'lastName', title: 'Last Name' },
+  { name: 'phone', title: 'Phone' },
+  { name: 'state', title: 'State' },
+];
+
+const requiredRule = {
+  isValid: value => value?.length > 0,
```
MaximKudriavtsev:
```js
isValid: value => value?.trim().length > 0
```

packages/dx-react-grid-demos/src/demo-sources/grid-editing/validation-on-typing.jsxt
```diff
@@ -0,0 +1,78 @@
...
+import debounce from 'lodash.debounce';
+
+import { generateRows, employeeValues } from '../../../demo-data/generator';
+
+const EditCell = ({ errors, ...props }) => {
+  const { children } = props;
+  return (
+    <TableEditColumn.Cell {...props}>
+      {React.Children.map(children, child => (
+        child && child.props.id === 'commit'
```
MaximKudriavtsev: child && child.props.id -> child?.props.id

ushkal:
> Validation On Typing demo
Should we highlight invalid cells (using border or background color)? It seems doesn't clear why SAVE is blocked.

I think we should keep this demo simple

> Should we make a new guide instead of a section? I think it will be better to find information and reduces guide sizes

From the other hand, this is not a builtin feature and the guide is too small. @RomanTsukanov what do you think?

RomanTsukanov:
As far as I see, the only instruction we can give is to handle onCommitChanges. If this is it, then there's indeed too little information for a guide. But if there are more instructions or use-cases to cover, we should probably make a separate article.

packages/dx-react-grid-demos/src/demo-sources/grid-editing/validation-on-typing.jsxt
```diff
@@ -0,0 +1,78 @@
...
+      {React.Children.map(children, child => (
+        child?.props.id === 'commit'
+          ? React.cloneElement(child, { disabled: errors[props.tableRow.rowId] })
+          : child
+      ))}
+    </TableEditColumn.Cell>
+  );
+};
+
+// maps the rows so that each prop is true if a column is required but the field is empty
```
RomanTsukanov:
```suggestion
// Maps the rows to a single object in which each field are is a row IDs
// and the field's value is true if the cell value is invalid (a column is required but the cell value is empty)
```

packages/dx-react-grid/docs/guides/editing.md
```diff
@@ -62,3 +62,13 @@ You can handle other keyboard actions. For example, the demo below shows how to
...
 
 .embedded-demo({ "path": "grid-editing/disable-column-editing", "showThemeSelector": true })
+
+## Data Validation
+
+Changes can be validated on-the-fly as shown in the example below. The 'Save' button gets disabled if any reqired field is empty.
+
+.embedded-demo({ "path": "grid-editing/validation-on-typing", "showThemeSelector": true })
+
+Another option is to validate data in the `onCommit` function. In this case it is possible to rollback invalid changes by omitting the `setState` call or display the error in a grid as in the demo below.
```
RomanTsukanov:
```suggestion
Changes can also be validated in the `onCommit` function. In this case, you can omit the `setState` call and rollback the changes if they are invalid.
```

packages/dx-react-grid/docs/guides/editing.md
```diff
@@ -62,3 +62,13 @@ You can handle other keyboard actions. For example, the demo below shows how to
 You can prevent editing of a specific column using the [EditingState](../reference/editing-state.md) plugin's `columnExtensions` property.
 
 .embedded-demo({ "path": "grid-editing/disable-column-editing", "showThemeSelector": true })
+
+## Data Validation
+
+Changes can be validated on-the-fly as shown in the example below. The 'Save' button gets disabled if any reqired field is empty.
```
RomanTsukanov:
```suggestion
Changes can be validated as the user types. You can disable the Save button if any required field is empty.
```

packages/dx-react-grid/docs/guides/editing.md
```diff
@@ -62,3 +62,13 @@ You can handle other keyboard actions. For example, the demo below shows how to
 You can prevent editing of a specific column using the [EditingState](../reference/editing-state.md) plugin's `columnExtensions` property.
```
dirk-pieterse:
```suggestion
You can use the [EditingState](../reference/editing-state.md) plugin's `columnExtensions` property to prevent a specific column for being edited.
```

PR #2695, opened 2020-02-21
refactor(react-scheduler): render All-Day Panel as part of the Timetable when vertical grouping is used

AryamnovEugeniy:


packages/dx-react-scheduler/src/plugins/all-day-panel.tsx
```diff
@@ -85,51 +88,88 @@ class AllDayPanelBase extends React.PureComponent<AllDayPanelProps, AllDayPanelS
...
+
+        <Template name="timeTable">
           {(params: any) => (
             <TemplateConnector>
-              {({ groupOrientation, currentView }) => {
-                if (currentView.type === VIEW_TYPES.MONTH) {
-                  return <TemplatePlaceholder params={{ ...params }} />;
+              {({ currentView, groupOrientation, allDayCellsData }) => {
+                if (currentView.type === VIEW_TYPES.MONTH
+                    || groupOrientation?.(currentView.name) !== VERTICAL_GROUP_ORIENTATION) {
```
LazyLahtak: `currentView.type` and `currentView.name`. 
These fields seems very similar.
Maybe we should rename 'currentView.name` => `currentView.orientation` or something like that?

AryamnovEugeniy: These fields serve different functions. `Type` is responsible for the view's type. Currently, there are three of them: 'month', 'week' and 'day'. `name`, however, is view's unique identifier. This is how we can differentiate between two or more view's of the same type (for example, there may be 5 month views)

packages/dx-scheduler-core/src/plugins/week-view/utils.ts
```diff
@@ -46,3 +47,15 @@ export const getLabelsForAllGroups: PureComputed<
     ) as TimeScaleLabelData[],
   ], [] as TimeScaleLabelData[][]);
 };
+
+export const prepareVerticalViewCellsData: PureComputed<
+  [ViewCell[][], AllDayCell[][]], ViewCell[][][]
+> = (cellsData, allDayCellsData) => {
+  const groupCount = allDayCellsData ? allDayCellsData.length : 1;
```
LazyLahtak: We can use
```jsx
const groupCount = allDayCellsData?.length || 1;
```

packages/dx-react-scheduler-material-ui/src/templates/grouping-panel/vertical-layout.jsx
```diff
@@ -25,19 +27,24 @@ const VerticalLayoutBase = ({
   classes,
   className,
   cellTextTopOffset,
+  alignWithAllDayRow,
   ...restProps
 }) => {
   const timeTableCellHeight = BASIC_CELL_HEIGHT[viewType];
+  const allDayCellHeight = BASIC_CELL_HEIGHT[VIEW_TYPES.ALL_DAY_PANEL];
```
MaximKudriavtsev: Seems that it is a constant value. Can we move it from `render` method?

packages/dx-react-scheduler-material-ui/src/templates/views/vertical/time-scale/ticks-layout.jsx
```diff
@@ -20,37 +20,60 @@ const TicksLayoutBase = ({
...
-      ))}
-    </TableBody>
-  </Table>
-);
+}) => {
+  const groupHeight = cellsData.length / groupCount;
+  return (
+    <Table {...restProps} className={classNames(classes.table, className)}>
+      <TableBody>
+        {cellsData.map((days, index) => (
```
MaximKudriavtsev:
```js
cellsData.map((days, index) => {});
// ->
cellsData.map(([firstDay], index) => {});
```

packages/dx-react-scheduler-material-ui/src/templates/views/vertical/time-table/layout.test.jsx
```diff
@@ -69,5 +70,38 @@ describe('Vertical view TimeTable', () => {
       expect(defaultProps.setCellElementsMeta)
         .toBeCalledTimes(2);
     });
+    it('should render timetable and al-day cells and rows', () => {
```
MaximKudriavtsev: al-day -> allDay or all-day ;)

packages/dx-react-scheduler/src/plugins/all-day-panel.tsx
```diff
@@ -85,51 +88,88 @@ class AllDayPanelBase extends React.PureComponent<AllDayPanelProps, AllDayPanelS
...
           </TemplateConnector>
         </Template>
 
+        <Template name="timeScale">
+          {(params: any) => (
+            <TemplateConnector>
+              {({ currentView, groupOrientation }) => {
+                if (currentView.type === VIEW_TYPES.MONTH
+                    || groupOrientation?.(currentView.name) !== VERTICAL_GROUP_ORIENTATION) {
+                  return <TemplatePlaceholder params={...params} />;
```
MaximKudriavtsev: This condition used three times. Can we extract it into one function?

packages/dx-react-scheduler/src/plugins/drag-drop-provider.tsx
```diff
@@ -312,13 +314,43 @@ class DragDropProviderBase extends React.PureComponent<
...
+            {({ currentView, groupOrientation }) => {
+              if (groupOrientation?.(currentView.name) === VERTICAL_GROUP_ORIENTATION) {
+                return (
+                  <>
+                    <TemplatePlaceholder />
+                    {renderAppointmentItems(this.allDayDraftAppointments, Container, draftData)}
+                    {renderAppointmentItems(this.timeTableDraftAppointments, Container, draftData)}
+                  </>
+                );
+              }
```
MaximKudriavtsev: Instead of `if` use ternary operator

```jsx
<>
  <TemplatePlaceholder />
  {renderAppointmentItems(this.allDayDraftAppointments, Container, draftData)}
  {groupOrientation?.(currentView.name) === VERTICAL_GROUP_ORIENTATION && renderAppointmentItems(this.timeTableDraftAppointments, Container, draftData)}
</>
```

And same places too.

MaximKudriavtsev:
DnD doesn't work with the AllDayPanel
Screen Recording 2020-03-12 at 10 34 36

MaximKudriavtsev:
image
I think this border would be more user-friendly

MaximKudriavtsev:
Note we should not forget to add docs and examples about Vertical Grouping. Maybe in the next PR

packages/dx-react-scheduler-material-ui/src/templates/views/vertical/time-scale/layout.jsx
```diff
@@ -85,6 +108,8 @@ export const Layout = ({
         cellsData={cellsData}
         className={classes.ticks}
         groupOrientation={groupOrientation}
+        groupCount={groupCount}
+        includeAllDayCell={showAllDayTitle}
```
MaximKudriavtsev: For future. We should not use specific names. I mean that this component calls TicksLayout and this component should not have information about grouping)

AryamnovEugeniy:
FPR

packages/dx-react-scheduler/src/plugins/all-day-panel.tsx
```diff
@@ -80,6 +84,10 @@ class AllDayPanelBase extends React.PureComponent<AllDayPanelProps, AllDayPanelS
     );
   });
 
+  allDayPanelExistsComputed = memoize(({
+    currentView,
+  }) => currentView.type !== VIEW_TYPES.MONTH);
```
ushkal: `isMonthView`?

packages/dx-react-scheduler/src/plugins/drag-drop-provider.tsx
```diff
@@ -346,13 +348,28 @@ class DragDropProviderBase extends React.PureComponent<
...
 
         <Template name="allDayPanel">
-          <TemplatePlaceholder />
-          {renderAppointmentItems(this.allDayDraftAppointments, Container, draftData)}
+          <TemplateConnector>
+            {({ currentView, groupOrientation }) => (
+              <>
+                <TemplatePlaceholder />
+                {groupOrientation?.(currentView.name) !== VERTICAL_GROUP_ORIENTATION
+                  && renderAppointmentItems(this.allDayDraftAppointments, Container, draftData)}
```
ushkal: I think it would be more clear if we use `if () {..} else return null;` approach

PR #2713, opened 2020-03-03
fix(scheduler-core): filter WeekView appointments correctly when excluded days contain an interval of three or more days in a row

### LINKED ISSUE #2712: Scheduler - excluded days prop may cause errors
### I'm using ...

- [ ] React Grid
- [ ] React Chart
- [x] React Scheduler

### Current Behaviour
When `excludedDays` is equal to [0, 5, 6] the Scheduler does not work

### Steps to Reproduce
https://codesandbox.io/s/devextreme-react-scheduler-for-material-ui-fq6fc

AryamnovEugeniy:
Fixes #2712

### DIFF
--- a/packages/dx-scheduler-core/src/utils.test.ts
+++ b/packages/dx-scheduler-core/src/utils.test.ts
@@ -12,6 +12,7 @@ import {
   filterByViewBoundaries,
   getRRuleSetWithExDates,
   formatDateToString,
+  excludedIntervals,
 } from './utils';
 
 describe('Utils', () => {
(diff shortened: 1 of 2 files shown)


packages/dx-scheduler-core/src/utils.test.ts
```diff
@@ -93,6 +94,48 @@ describe('Utils', () => {
...
+    it('should create an interval for several excluded days in a row', () => {
+      const excludedDays = [2, 3, 4, 5];
+
+      expect(excludedIntervals(excludedDays, start))
+        .toEqual([[
+          moment(start.day(2)),
+          moment(start.day(5)).endOf('day'),
+        ]]);
+    });
+    it('should work correctly with sundays', () => {
```
MaximKudriavtsev: What difference between `sundays` and basic days?

Maybe call this test something like that - `should work with week boundaries`

PR #2715, opened 2020-03-03
fix(react-scheduler): make it possible to change allowDrag and allowResize in runtime

### LINKED ISSUE #2710: Scheduler - It is impossible to change allowDragging and allowResizing in runtime

### I'm using ...

- [ ] React Grid
- [ ] React Chart
- [x] React Scheduler

### Steps to Reproduce
1) Open https://codesandbox.io/s/ancient-monad-rtpwp
2) Try dragging - it works
3) Disable dragging - dragging still works, though it shouldn't

Same with `allowResizing`

AryamnovEugeniy:
Fixes #2710


packages/dx-react-scheduler/src/plugins/drag-drop-provider.tsx
```diff
@@ -67,6 +72,27 @@ class DragDropProviderBase extends React.PureComponent<
...
+    }
+
+    return {
+      ...state,
+      appointmentContentTemplateKey:
+        isAllowDragUpdated ? state.appointmentContentTemplateKey : Math.random(),
+      appointmentTopTemplateKey:
+        isAllowResizeUpdated ? state.appointmentTopTemplateKey : Math.random(),
+      appointmentBottomTemplateKey:
+        isAllowResizeUpdated ? state.appointmentBottomTemplateKey : Math.random(),
```
MaximKudriavtsev: Where do you set `allowDrag` and `allowResize` state properties?

packages/dx-react-scheduler/src/plugins/drag-drop-provider.test.tsx
```diff
@@ -288,6 +288,70 @@ describe('DragDropProvider', () => {
       expect(allowResize)
         .toBeCalledWith(deps.template.appointmentTop.data);
     });
+    it('should change Resize render in runtime depending on allowResize prop', () => {
+      const allowResizing = () => true;
+      const deps = {
```
MaximKudriavtsev: Does it necessary for the test?

AryamnovEugeniy: Yes. We need these templates, otherwise they wouldn't exist and Resize component wouldn't be rendered at all

PR #2748, opened 2020-03-16
fix(react-grid): reset viewport position on filter change when data remote

### LINKED ISSUE #2739: VirtualRowsCache not reset when onFiltersChange in

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour
When using VirtualTable with remote filtering, if one scrolls further down resulting in loading of more data, if one filters by a value that is clearly in the table (meaning it exists), just an empty grid with the right column count is rendered.

After some investigation, I believe it's because the VirtualRowsCache is not reset when the filter parameters change which leaves `skip` value as the same.  This results in wrong parameters being sent while fetching remote data. e.g
if skip = 200 and take=300, if the result is among the first 10 results, it is not returned and yet the totalRowCount will be correct. 

### Expected Behaviour
The `skip` value should be reset to 0 when onFiltersChange in the FilteringState is called. 

### Steps to Reproduce
While using `VirtualTableBase` and FilteringState with remote filters i.e filters and onRemoteFiltering managed externally:
- Scroll down until more results are loaded in the table. i.e `skip` > 0
- Filter by any value currently visible in the grid

### What I have tried
One way I can solve this by directly calling `clearRowCache` action. This can be through:
- extending FilteringState and using a custom plugin with TemplateConnector and calling actions.clearRowCache(). e.g
```
<Plugin>
    <TemplateConnector>
      {(getters, actions) => {
        actions.clearRowCache();
        return null;
      }}
    </TemplateConnector>
  </Plugin>
```
The problem with this is it's called in render hence problems of updating state while rendering.
- Calling actions.clearRowCache before I call `onFiltersChange`, but there is no way (that I know of) to access the global `actions` without going though a plugin with TemplateConnector like above.

### Screenshots
Screen Shot 2020-03-12 at 1 28 22 am

### Environment

- devextreme-reactive: 19.1.6
- devextreme-react: 19.1.6
- dx-react-grid-bootstrap4: 2.0.5
- react: 16.8.1
- browser: chrome (all browsers though)
- bootstrap: 4

LazyLahtak:
Fixes #2739


MaximKudriavtsev:
Can we add some tests for this case? I think it would be very helpful in the future =)

LazyLahtak:
> Can we add some tests for this case? I think it would be very helpful in the future =)

I'll add it after rechecks this fix. @AryamnovEugeniy possibly found a way to reproduce this bug after fix.

MaximKudriavtsev:
@LazyLahtak great!

packages/dx-react-grid/src/plugins/virtual-table/virtual-table-state.tsx
```diff
@@ -55,9 +55,14 @@ class VirtualTableStateBase extends React.PureComponent<VirtualTableStateProps,
     const { requestedRange, actualBounds } = getRequestMeta(
       referenceIndex, virtualRows, pageSize!, totalRowCount, forceReload, infiniteScrolling,
     );
-
-    if (forceReload || shouldSendRequest(requestedRange, requestedStartIndex)) {
-      this.requestNextPage(requestedRange, actualVirtualRows, actualBounds);
+    const isRangeFinite = isIntervalFinite(requestedRange);
+
+    if (forceReload || !isRangeFinite || shouldSendRequest(requestedRange, requestedStartIndex)) {
+      if (isRangeFinite) {
```
ushkal: `range = isRangeFinite ? requestedRange : ZERO_INTERVAL` would be better

LazyLahtak:
Ready to review

packages/dx-react-grid/docs/guides/lazy-loading.md
```diff
@@ -30,8 +30,6 @@ Row caching helps reduce the number of data requests. Call the [createRowCache](
 
 If users are allowed to sort or filter rows, clear the row cache each time the sort or filter settings are changed. Filtering also requires the total number of rows to be returned in every response.
 
-*NOTE: Debounce requests on filter change to avoid fetching data on each keystroke.*
```
ushkal: We should keep this note

LazyLahtak: But we don't use debounce in Demo.

ushkal: we should warn users nevertheless

PR #2753, opened 2020-03-17
fix(react-grid): make 'onTopRowChange' work with detail row

### LINKED ISSUE #2743: React Grid crashes when opening the last TableRowDetail row while using VirtualTable

### I'm using ...
- [x] React Grid

### Current Behaviour
I am running into a bug where the Grid crashes when opening the last TableRowDetail and scrolling down if a VirtualTable is used. This error is thrown:

Uncaught TypeError: Cannot read property 'rowId' of undefined at VirtualTable._this.setViewport (dx-react-grid.es.js:3705)

This is caused by this function
`
    setViewport = (viewport, { tableBodyRows, isDataRemote }: Getters) => {
      const { onTopRowChange } = this.props;
      const hasViewportRows = viewport && viewport.rows;
      const hasBodyRows = tableBodyRows && tableBodyRows.length;
      const rowId = hasViewportRows && hasBodyRows && !isDataRemote
        ? tableBodyRows[viewport.rows[0]].rowId
        : undefined;

      onTopRowChange(rowId);
      this.setState({ viewport });
    }
`

viewport.rows[0] contains a value which is bigger than the length of the tableBodyRows array. tableBodyRows[viewport.rows[0]] therefore results in undefined, hence the error thrown when trying to read rowId from undefined.

This only seems to occur when the TableRowDetail is rather tall.

### Link to Sandbox: 
https://codesandbox.io/s/dawn-hooks-sb7i2

### Steps to Reproduce

Scroll down to the last item in the grid.
Open the detail row.
Try to scroll down again
The error should occur.
I hope you can provide us with a fix soon, because this is rather urgent for us.

Thanks in advance.

LazyLahtak:
Fixes #2743

### DIFF
--- a/packages/dx-grid-core/src/plugins/virtual-table/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/virtual-table/helpers.test.ts
@@ -1,4 +1,10 @@
-import { getViewport, checkColumnWidths, calculateScrollHeight, getScrollTop } from './helpers';
+import {
+  getViewport,
+  checkColumnWidths,
+  calculateScrollHeight,
+  getScrollTop,
+  getTopRowId,
+} from './helpers';
 import { TOP_POSITION, BOTTOM_POSITION } from './constants';
 
 const estimatedRowheight = 40;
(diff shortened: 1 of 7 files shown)


packages/dx-grid-core/src/plugins/virtual-table/helpers.test.ts
```diff
@@ -211,4 +217,43 @@ describe('#checkColumnWidths', () => {
         .toEqual(rowHeight * rows.length);
     });
   });
+
+  describe('#getTopRowId', () => {
+    const rows = [{ rowId: 1 }, { },  { rowId: 2 }, { rowId: 3 }, { rowId: 4 }, { rowId: 5 }];
+
+    it('should work', () => {
+      const viewport = { rows: [0, 0] };
```
AryamnovEugeniy: We may call it `viewportBase` and use it in other tests as `rows`

Issue #2656, opened 2020-01-29
VirtualTable crashes with the "Maximum update depth exceeded" error 


AlexanderKovalenko:
The issue occurs in the following code example when the wordWrapEnabled option is enabled:

[DevExtreme React Grid for Bootstrap 4 - CodeSandbox](https://codesandbox.io/s/devextreme-react-grid-for-bootstrap-4-454yl)

Original Support Center ticket:

https://supportcenter.devexpress.com/ticket/details/t856671/grid-crashes-when-the-virtualtable-plugin-is-used

Issue #2550, opened 2019-11-20
Material-ui tab, not all columns appear


ahmetcaglayan:
https://codesandbox.io/s/shy-snowflake-z3d0p

material-ui tab doesn't show all the columns of the table I'm using. 
The problem is the second menu. 
When I put the table in the first menu does not occur. 
But when I switch to the second menu, the problem occurs, it only shows the first 2 columns. 
Cannot make full width adjustment.

https://material-ui.com/components/tabs/

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

LazyLahtak:
Hi @ahmetcaglayan,

Thank you for your report. We need additional time to research this behaviour.

LazyLahtak:
Hi @ahmetcaglayan,

We found that that fix of this issue breaks the number of other popular scenarios and affect a lot of users.  So, decided to revert it in this PR here: #2771.

We also double-checked this particular case and found you can overcome the issue using this approach: https://codesandbox.io/s/keen-ride-uojre

PR #2771, opened 2020-03-23
fix(react-grid): prevent infinite loop while horizontal scroll

### LINKED ISSUE #2770: Reactive Grid - The "Maximum update depth exceeded" error occurs on an attempt to horizontally scroll the grid when the size of the rows differs

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

The "Uncaught Error: Maximum update depth exceeded." error occurs on an attempt to use the horizontal scroll in the grid

### Expected Behaviour

No errors occur

### Steps to Reproduce

1) Open the following example:
https://codesandbox.io/s/devextreme-react-grid-for-material-ui-0dvvj

2) Try to scroll the grid to the right and left

### Screenshots

image

### Environment

The issue is reproducible starting with v2.3.1. In v2.3.0 and earlier it all works fine.

- devextreme-reactive: 2.5.1
- react: 16.13.0
- browser: Chrome, FF
- bootstrap: none
- react-bootstrap: none
- material-ui: 4.9.5

LazyLahtak:
Fixes #2770

Fix regression after:
* #2577 (fix issue  #2550)
* #2662 (fix issue #2656)

Issue #2550 resolves by using `key` property in the `Grid` component. [Example](https://codesandbox.io/s/keen-ride-uojre)
Issue #2656 caused by #2577.


MaximKudriavtsev:
How do we test this behavior?

And what about the following comment?
> NOTE: Original Sizer doesn't work correctly with Material-UI Tabs (#2550)

PR #2800, opened 2020-04-07
docs: Make minor changes to the Export to Excel articles

RomanTsukanov:


packages/dx-react-grid/docs/guides/export.md
```diff
@@ -50,7 +50,7 @@ Note that the [plugin order](./plugin-overview.md#plugin-order) is important.
 
 4. Save the file
 
-   To handle file saving, implement the `onSave` callback of the `GridExporter` component. In this callback, save the file to the user's local storage. This can be done in many different ways. For example, you can use the `file-saver` package:
+   To save the file to the user's local storage, implement the `onSave` callback of the `GridExporter` component. In the following code, the file is saved using the `file-saver` package, but you can use any other package that provides similar functionality.
```
albertov05, line 53: Can we instead say:
To save the file to local storage at runtime, use the `onSave` callback for the `GridExporter` component. In the following code, the file is saved via the `file-saver` package, but you can use any other similar package.

packages/dx-react-grid/docs/guides/export.md
```diff
@@ -68,7 +68,7 @@ The following demo shows export in action:
 
 ## Supported Grid Features
 
-Exported data can be transformed in the Excel document as it is in the Grid. For this, pass Grid settings to the following `GridExporter` properties:
+Data shaping settings applied to the Grid can be also applied to the exported Excel document. For this, pass Grid settings to the following `GridExporter` properties:
```
albertov05, line 71: Can we say:
To apply a grid's data shaping settings to an exported Excel document, pass the settings to the following `GridExporter` properties:

packages/dx-react-grid/docs/guides/export.md
RomanTsukanov: It's important that the local storage is the user's, not the developer's. In addition, the phrase "use the callback for the component" doesn't sound right. "**Implement** the callback for the component" does, so I suggest the following version:

```suggestion
   To save the file to the user's local storage at runtime, implement the `onSave` callback for the `GridExporter` component. In the following code, the file is saved via the `file-saver` package, but you can use any other similar package.
```

packages/dx-react-grid/docs/guides/export.md
RomanTsukanov:
```suggestion
To apply the Grid's data shaping settings to an exported Excel document, pass the settings to the following `GridExporter` properties:
```

packages/dx-react-grid/docs/guides/export.md
albertov05: I see. Then let's change it to:
To save a file to a user's local storage at runtime, implement the `onSave` callback for the `GridExporter` component.

PR #2879, opened 2020-05-08
fix(react-grid): render drag drop container in body

### LINKED ISSUE #2853: Drag-n-drop grouping working incorrectly in container with attribute 'transform: translate(x, y)'

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour
I'm using React Grid with grouping panel in popup window and that window positioned on page with css property 'transform; translate(x, y)'. When I'm trying to drag and drop column from table to grouping panel the dragged component located not exactly on cursor. Difference beetween correct position and current the same as values in **translate(x, y)**. If I'm turn of transform attribute all works as expected.

### Expected Behaviour
Dragged component located exactly under the cursor.

### Steps to Reproduce
1. Open sandbox example
2. Drag and drop column to Grouping panel

Link to the example that reproduces the issue: &lt;https://codesandbox.io/s/elegant-kilby-li5sl?file=/demo.tsx&gt;

### Environment
Reproduced in devextreme's example 
https://codesandbox.io/s/elegant-kilby-li5sl?file=/demo.tsx

LazyLahtak:
Fixes #2853


packages/dx-react-grid/src/plugins/drag-drop-provider.tsx
```diff
@@ -53,18 +54,21 @@ class DragDropProviderBase extends React.PureComponent<DragDropProviderProps, Dr
...
+                    {getTargetColumns(payload, columns)
+                      .map(column => (
+                        <Column
+                          key={column.name}
+                          column={column}
+                        />
+                      ))
+                    }
+                  </Container>,
+                  document.body,
```
MaximKudriavtsev, line 70: This code doesn't work on SSR. We should avoid using the `document` variable in the render method.

LazyLahtak:
Ready to review

PR #2973, opened 2020-07-06
fix(react-grid): improve fixed columns reordering

### LINKED ISSUE #2944: Fixed columns dosen't work with Drag and Drop
### I'm using ...

- [x] React Grid
- [x] Bootstrap4

### Current Behaviour
When TableFixedColumns are used at the same time as TableColumnReordering by the DragDropProvider, it is neither intuitive or practical what happens. 

Initial state: If you have multiple columns (contains more space than the component provides, so you have a horizontal scrollbar), and you slide to the right, when your leftmost columns are fixed to the left side.

Problem: If I have two columns fixed in the left side, and I try to take the rightmost (of the fixed) and put it to the left of the other fixed column this wont work. It seems like the drag 'n drop feature will only look at the columns hidden in the background instead of the fixed position. 

### Expected Behaviour
It would be intuitive and practical to be able to rearange the fixed columns. The drag and drop feature should look at the fixed columns at first, then if and only if there is non nearby go ahead and look at the regular columns as it is now.

### Link to reproduction
Link to the example that reproduces the issue: [link](https://codesandbox.io/s/proud-darkness-5c3r0)
Try to scroll to the right, and move `Channel` to the left of `Region`

LazyLahtak:
Fixes #2944


packages/dx-grid-core/src/utils/column-geometries.test.ts
```diff
@@ -55,4 +55,75 @@ describe('ColumnGeometries utils', () => {
...
+
+      const secondNode = { getBoundingClientRect, style: { left: '200px' } };
+      expect(getCellGeometries(secondNode)).toEqual({
+        isFixed: true,
+        left: 200,
+        right: 300,
+      });
+    });
+
+    it.only('should return dimensions based on "right" style value', () => {
```
AryamnovEugeniy, line 88: Do we need it here?

PR #2979, opened 2020-07-09
fix(react-grid): make TableBandHeader work correctly with TableColumnVisibility

### LINKED ISSUE #2963: Grid - The table has incorrect right border after hiding a data column
### I'm using ...
- [x] Reactive Grid

**Issue:**

The table has incorrect appearance after hiding a data column

**Example** 
https://codesandbox.io/s/naughty-torvalds-6454h?file=/demo.js:2361-2376

**Steps:** 

Open the example and hide the Area column using ColumnChooser

**Actual results:**

The last column's right border disappears

**Expected results:**

The last column should have the right border

**Additional info:**

The problem is related to the plugin order and doesn't appear if `TableColumnVisibility` is defined *before* `TableBandHeader`
Example - https://codesandbox.io/s/musing-star-j83gi?file=/demo.js

LazyLahtak:
Fixes #2963

BREAKING CHANGE:

The `TableColumnVisibility` plugin must now be declared **before** the `TableBandHeader` plugin. This is required to correctly calculate column spans and borders.

```diff
...
+<TableColumnVisibility
+  defaultHiddenColumnNames={...}
+/>
...
<TableBandHeader
  columnBands={columnBands}
/>
...
-<TableColumnVisibility
-  defaultHiddenColumnNames={...}
-/>
...
```

### DIFF
--- a/packages/dx-grid-core/src/plugins/table-band-header/helpers.test.ts
+++ b/packages/dx-grid-core/src/plugins/table-band-header/helpers.test.ts
@@ -232,7 +232,7 @@ describe('TableBandHeader Plugin helpers', () => {
         });
     });
 
-    it('should return a null-typed band component for column without key', () => {
+    it('should return a an empty cell type for column without key', () => {
       const params = {
         tableColumn: {
           type: TABLE_DATA_TYPE,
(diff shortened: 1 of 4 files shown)


packages/dx-grid-core/src/plugins/table-band-header/helpers.test.ts
```diff
@@ -232,7 +232,7 @@ describe('TableBandHeader Plugin helpers', () => {
         });
     });
 
-    it('should return a null-typed band component for column without key', () => {
+    it('should return a an empty cell type for column without key', () => {
```
AryamnovEugeniy, line 235: should return an empty cell type for column without key

PR #2999, opened 2020-07-20
fix(react-core): make scrolling behavior similar in Safari and Chrome while drag columns

### LINKED ISSUE #2982: Prevent horizontal scrolling during drag operation

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

Grid horizontally scrolls when we hold on the dragged column and move to the extreme right.

### Expected Behaviour

Prevent horizontal scrolling during drag operation

### Steps to Reproduce

It only happens in safari browser. Drag the first column to the extreme right and you will see grid automatically scrolls towards right side. We don't see this behaviour in other browsers (Chrome, Firefox). We want to prevent this scrolling even in case of safari browser.

https://codesandbox.io/s/drag-and-drop-17q1o

### Environment

- devextreme-reactive: 2.6.4
- react: 16.13.1
- browser: Safari 13.0.2
- bootstrap: none
- react-bootstrap: none
- material-ui: 4.10.2

MaximKudriavtsev:
Fixes #2982

### DIFF
--- a/packages/dx-react-core/src/draggable.tsx
+++ b/packages/dx-react-core/src/draggable.tsx
@@ -75,12 +75,13 @@ export class Draggable extends React.Component<DraggableProps> {
     if (!node) return;
     node.removeEventListener('mousedown', this.mouseDownListener);
     node.removeEventListener('touchstart', this.touchStartListener);
-    node.addEventListener('mousedown', this.mouseDownListener, { passive: true });
+    node.addEventListener('mousedown', this.mouseDownListener);
     node.addEventListener('touchstart', this.touchStartListener, { passive: true });
   }
 
   mouseDownListener(e) {
     if (this.touchStrategy.isWaiting() || e[draggingHandled]) return;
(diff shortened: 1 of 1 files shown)


MaximKudriavtsev:
@LazyLahtak @AryamnovEugeniy 

I am not sure, why there was a `{ passive: true }` flag in [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). Do you know why?

LazyLahtak:
> @LazyLahtak @AryamnovEugeniy
> 
> I am not sure, why there was a `{ passive: true }` flag in [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). Do you know why?

As I found, listener with `{ passive: true }` ignores `preventDefault()`.
Probably, we should get rid of the `{ passive: true }` or the `preventDefault()`.

PR #3000, opened 2020-07-21
docs(react-grid): Row reordering guide

LazyLahtak:
[Simple Grid Demo](https://codesandbox.io/s/jolly-pine-w8913)
[Tree Data Demo](https://codesandbox.io/s/upbeat-cache-cxgcq)


packages/dx-react-grid/docs/guides/third-party-libraries.md
```diff
@@ -0,0 +1,45 @@
...
+  <Plugin
+    name="OrderedRowProvider"
+    dependencies={[{ name: 'CustomTreeData' }]}
+  >
+    <Getter name="rows" computed={handleOrderedRows(handleRows)} />
+  </Plugin>
+);
+```
+
+Secondly, when you use [Tree Data](./tree-data.md) each row should be able to change parent. Also, you in some cases you have row array without parents. In this case, you may need to make a one row the children of another row. In demo below, you can do this by click on reordering control:
```
MaximKudriavtsev, line 44:
```suggestion
Secondly, when you use [Tree Data](./tree-data.md) each row should be able to change parent. Also, in some cases, you have a row array without parents. In this case, you may need to make one row the children of another row. In the demo below, you can do this by click on reordering control:
```

packages/dx-site/page-navigation.json
```diff
@@ -147,6 +147,10 @@
             "title": "TypeScript Support",
             "path": "/react/grid/docs/guides/typescript/"
           },
+          {
+            "title": "Third-party libraries",
```
MaximKudriavtsev, line 151: The Row Reordering is the better name for this topic

packages/dx-react-grid/docs/guides/third-party-libraries.md
MaximKudriavtsev: I don't understand why this paragraph is here

packages/dx-react-grid-demos/src/demo-sources/grid-row-reordering/tree.jsxt
```diff
@@ -0,0 +1,163 @@
...
+
+    const source = orderedRows[oldIndex];
+    const sourceIndex = data.indexOf(source);
+    const sourceParentId = source.parentId;
+
+    const target = orderedRows[newIndex];
+    const targetIndex = data.indexOf(target);
+    const targetParentId = target.parentId;
+
+    let newData;
```
AryamnovEugeniy, line 89: It is better to move newData calculations outside render method (to leave useCallback here, but to create a separate function which would calculate newData based on `data`).

PR #3066, opened 2020-08-27
adds the date from each individual cell into appointment.dataItem

agathayin:
The main purpose of this change is to get the accurate date of the cell clicked by the user. 
In dx-scheduler-core/src/plugins/horizontal-rect & vertical-rect, date for individual cell is cloned into appointment.dataItem.cellRectDate so that when AppointmentTooltip is opened, both tooltip header and container have a prop showing the accurate date of the chosen cell. 

This function can be used for creating custom functions on the tooltip. With the accurate date, new features such as editing or deleting only one day from repeating appointments can be implemented as buttons on Appointment Tooltip.

I wasn't able to pass the automatic checks but the codes work when I am using the scheduler.
Screen Shot 2020-08-27 at 3 43 31 PM

### DIFF
--- a/packages/dx-scheduler-core/src/plugins/horizontal-rect/helpers.ts
+++ b/packages/dx-scheduler-core/src/plugins/horizontal-rect/helpers.ts
@@ -8,6 +8,12 @@ const CELL_BOUND_OFFSET_PX = 1;
 const getCellRect: GetCellRectHorizontalFn = (
   date, appointment, viewCellsData, viewMetaData, cellElementsMeta, takePrev, multiline,
 ) => {
+  // Adds the date from each individual cell
+  // This is so I can get the date after selecting a cell
+  if((appointment as any).dataItem){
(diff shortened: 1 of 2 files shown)


AryamnovEugeniy:
Hello!

Thank you for your interest in our product!
About this Pull Request. `getCellRect` function calculates the cell coordinates based on date and groups. It should not do anything else. Nor should it change any of its parameters because we try to make all of our functions pure (and you change `dataItem`, which makes this function not pure).
But the main thing is that the purpose of this PR is not clear. Could you describe it in greater detail? Maybe, we can find a solution which will not require changes in our code. For instance, it is possible to delete only one day from a recurring appointment through our tooltip (have a look at this demo: https://devexpress.github.io/devextreme-reactive/react/scheduler/demos/featured/overview/)

AryamnovEugeniy:
I'm closing this due to inactivity

Issue #3124, opened 2020-10-12
Typescript Typing issue for DataTypeProvider.ValueEditorProps


las3r:
- [x] I have searched this repository's [issues](https://github.com/devexpress/devextreme-reactive/issues) and believe that this is not a duplicate.

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

 `DataTypeProvider.ValueEditorProps` does not contain the proper definition for the actual values being passed.
 The below "signature" is what is being linked from your docs and it throws errors from typescript saying that `onBlur` and `autoFocus` are not part of the type.

 ```ts
const Component=  { value, onValueChange, onBlur, autoFocus }: DataTypeProvider.ValueEditorProps) = (<></>);
 ```
`Property 'onBlur' does not exist on type 'ValueEditorProps'.  TS2339`
`Property 'autoFocus' does not exist on type 'ValueEditorProps'.  TS2339`

 The following code works during runtime (the props are all passed and usable)
  ```ts
 const Component= ({ value, onValueChange, onBlur, autoFocus }: any) = (<></>);
 ```

Source: `dx-react-grid.d.ts`
Interface:
```ts
interface ValueEditorProps {
    /** A column object. */
    column: Column;
    /** A row. */
    row?: any;
    /** Specifies the editor value. */
    value: any;
    /** Handles value changes. */
    onValueChange: (newValue: any) => void;
    /** "true" if users should not be able to edit the value, "false" otherwise. */
    disabled: boolean;
  }
```
### Expected Behaviour
Working typings :-)

### Environment

- devextreme-reactive: 2.7.2

LazyLahtak:
Hi @las3r,

Thank you for your interest in our Grid and for your [pull request](#3128).
There are some Breaking Changes in your pull request: you removed some types that already exist (some other customer could use them). For this reason we cannot accept these changes. Also, all types should be added in [documentation](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-grid/docs/reference/data-type-provider.md#datatypeprovidervalueeditorprops).
However, I'll be happy to help you to complete and merge your pull request.

PR #3128, opened 2020-10-13
fix(react-grid): fix typings of ValueEditorProps (#3124)

las3r:
This PR corrects the typing issue described in #3124


LazyLahtak:
Fixed in #3146

PR #3294, opened 2021-03-29
Update root.jsx

mathiaslima:


AryamnovEugeniy:
Hello!
We cannot change the currently existing layout for DateNavigator's Root because this would cause a Breaking Change. But this is possible from your side: you only need to override DateNavigator's rootComponent (DateNavigator has a private property called `rootComponent`). This will allow you to change the order of buttons in the DateNavigator.

Please, let me know if this solution suits you

PR #3358, opened 2021-06-29
fix: Do not treat right mouse click as the start of dragging (T1005625)

dxvladislavvolkov:


packages/dx-react-core/src/draggable.tsx
```diff
@@ -79,8 +80,12 @@ export class Draggable extends React.Component<DraggableProps> {
     node.addEventListener('touchstart', this.touchStartListener, { passive: true });
   }
 
+  isRightButton(mouseButton) {
+    return mouseButton === RIGHT_MOUSE_BUTTON;
```
LazyLahtak, line 84: This method doesn't use `this`. Should it be method or just utility function?

dxvladislavvolkov: I will add just a function

LazyLahtak (approved):

PR #3360, opened 2021-06-30
chore: support Material-UI 5

LazyLahtak:


fareeha-arshad:
#3378 

This pull request contains support for material UI v5 whose beta version has been out since July so please make this fix available as soon as possible so that dev extreme reactive library can be utilised with material ui v5.

benediktdertinger:
Hi guys, can you please give an update on when this will be released?

PR #3366, opened 2021-07-09
style: Add active button style user prop to grid

threepears:
As I have seen a few requests for it and the functionality doesn't seem to exist yet, I have created a prop named `activeButtonClass` that can be passed into the Grid when creating that component to let the user manipulate the CSS styles of the "active" button in the `PagingPanel`.  This prop will take a string that should be a user-generated class name, and will add it to the classes for a button with the `activeButton` class.  An example of how this would work in a React-Material UI setup is as follows:

```
import React, { useState } from 'react';
import Paper from '@material-ui/core/Paper';
import {
  PagingState,
  IntegratedPaging,
} from '@devexpress/dx-react-grid';
import {
  Grid,
  Table,
  TableHeaderRow,
  PagingPanel,
} from '@devexpress/dx-react-grid-material-ui';
import { makeStyles } from '@material-ui/core/styles'
import { generateRows } from '../../../demo-data/generator';

const useStyles = makeStyles((_theme) => ({
  myClass: {
    backgroundColor: 'maroon',
    color: 'white',
  },
}))

export default () => {
  const [columns] = useState([
    { name: 'name', title: 'Name' },
    { name: 'gender', title: 'Gender' },
    { name: 'city', title: 'City' },
    { name: 'car', title: 'Car' },
  ]);
  const [rows] = useState(generateRows({ length: 8 }));
  const [currentPage, setCurrentPage] = useState(0);
  const [pageSize, setPageSize] = useState(5);
  const [pageSizes] = useState([5, 10, 15]);

  const classes = useStyles()

  return (
    <Paper>
      <Grid
        rows={rows}
        columns={columns}
      >
        <PagingState
          currentPage={currentPage}
          onCurrentPageChange={setCurrentPage}
          pageSize={pageSize}
          onPageSizeChange={setPageSize}
        />
        <IntegratedPaging />
        <Table />
        <TableHeaderRow />
        <PagingPanel
          pageSizes={pageSizes}
          activeButtonClass={classes.myClass}
        />
      </Grid>
    </Paper>
  );
};
```

That is modified code from your `controlled-mode.g.jsx` file, and the result looks like this ....

sample-grid

There are still some issues to be addressed by the maintainers, however, and I can tweak accordingly.  Questions off the top of my head ...

1.  Is `activeButtonClass` an acceptable name for this prop?
2. I didn't yet create a test, because I wasn't sure exactly what type of test you would like (just that the class makes it to the button?) or where it should be placed.  Should I place something in `packages/dx-react-grid-material-ui/src/templates/paging-panel/pagination.test.jsx`?
3. As the styling is much better for Bootstrap (and as I only needed this for my Material UI setup ... as did most of the people asking about this feature in my online searches), I first fully implemented this prop for the `dx-react-grid-material-ui` module.  Once I had a test or two fail, I implemented for Bootstrap 4 .... Bootstrap 3 proved trickier, so I just handled the new prop without passing it through.  Is it critical to implement for the Bootstrap 3 module ... will this PR not be accepted if it is not included?

Anyway, please review and let me know if there are any further issues to address or changes to make!  I think this would be a useful feature to have for many people, so I'm hoping we can get this into the repo!  Thank you!

### DIFF
--- a/packages/dx-react-grid-bootstrap3/src/templates/paging-panel/pager.jsx
+++ b/packages/dx-react-grid-bootstrap3/src/templates/paging-panel/pager.jsx
@@ -5,6 +5,7 @@ import { PageSizeSelector } from './page-size-selector';
 import { Pagination } from './pagination';
 
 export const Pager = ({
+  activeButtonClass,
   currentPage,
   onCurrentPageChange,
   totalPages,
@@ -26,12 +27,12 @@ export const Pager = ({
     {...restProps}
   >
     {!!pageSizes.length && (
-    <PageSizeSelector
-      pageSize={pageSize}
-      onPageSizeChange={onPageSizeChange}
-      pageSizes={pageSizes}
-      getMessage={getMessage}
-    />
+      <PageSizeSelector
+        pageSize={pageSize}
+        onPageSizeChange={onPageSizeChange}
+        pageSizes={pageSizes}
+        getMessage={getMessage}
+      />
     )}
     <Pagination
       totalPages={totalPages}
@@ -40,6 +41,7 @@ export const Pager = ({
       onCurrentPageChange={page => onCurrentPageChange(page)}
       pageSize={pageSize}
       getMessage={getMessage}
+      activeButtonClass={activeButtonClass}
     />
   </div>
 );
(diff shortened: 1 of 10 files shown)


lcanavesio:
@LazyLahtak  could you check the code?

threepears:
Is there something else I need to do here?  Any procedures I have failed to take into account to get this considered?  Would love to get any movement on this one way or another .... thank you!

lcanavesio:
@threepears For now, until it is approved, I created my own component based on theirs and using my style and it worked perfectly.

Krijovnick:
Hi @threepears ,

Thank you for your interest in our products.

With your approach, it will not be possible to cover all possible user cases. We have no plans to implement this feature. You can find our roadmap here: https://js.devexpress.com/Roadmap.

You can override the `containerComponent` property to customize the entire Paging Panel and provide your own component based on our [code](https://github.com/DevExpress/devextreme-reactive/tree/master/packages/dx-react-grid-material-ui/src/templates/paging-panel).

Issue #3313, opened 2021-04-09
React Grid - Extra space is displayed towards the right side


onebalaban:
- [x] I have searched this repository's [issues](https://github.com/devexpress/devextreme-reactive/issues) and believe that this is not a duplicate.

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

Extra space is displayed towards the right side:

### Expected Behaviour

Grid should not display it.

### Steps to Reproduce

[CodeSandbox](https://codesandbox.io/s/damp-mountain-phnqv?file=/demo.js)

### Screenshots

<img width="646" alt="Screen Shot 2021-04-09 at 17 12 50" src="https://user-images.githubusercontent.com/25555034/114193116-e13c6180-9956-11eb-920b-5d79845d6864.png">

### Environment

- devextreme-reactive: 2.7.4
- react: 16.14.0
- browser: Chrome 89
- bootstrap: none
- react-bootstrap: none
- material-ui: 4.11.2

### Notes
T987831

PR #3471, opened 2021-12-30
fix(react-grid): fix colspan calculation

### LINKED ISSUE #3467: colspan wrongly set in group summaries cell,  which cause incorrect alignment

### I'm using ...

- [x] React Grid
- [ ] React Chart
- [ ] React Scheduler

### Current Behaviour

The "Sale Amount" align incorrectly

### Expected Behaviour

The summaries cell align to right column

### Steps to Reproduce

1. Set `showColumnsWhenGrouped` on `TableGroupRow`
2. Drag the column to the grouping panel, and the column is placed after a summarized column (here,  "units" is summarized , so we drag the "customer" to the grouping panel )
3. Expand group  row

sandbox link:  https://codesandbox.io/s/infallible-almeida-zm6ky?file=/demo.js

### Screenshots

image

### Environment

- devextreme-reactive: 2.7.6

### Description

Seems `tableGroupCellColSpanGetter` wrongly calculate,  because there are  2 `groupSummaryChain` started with 'custom' columnName,  one is for group type column,  another is for data type column.  But `chain` final set to `undefined` by `dataColumnGroupedBy` condition

Krijovnick:
Fixes #3467, #3313

### DIFF
--- a/packages/dx-grid-core/src/plugins/table-group-row/computeds.ts
+++ b/packages/dx-grid-core/src/plugins/table-group-row/computeds.ts
@@ -86,6 +86,7 @@ const groupSummaryChains: GroupSummaryChainsFn = (
   return sortAndSpliceColumns(tableColumns, firstVisibleColumnIndex)
     .reduce((acc, col) => {
       const colName = (col.column && col.column.name) as string;
+      const colKey = col.key;
       const isStartOfGroupCaption = col.type === TABLE_GROUP_TYPE
         && tableRow.row.groupedBy === colName;
       const isIndentColumn = col.type === TABLE_GROUP_TYPE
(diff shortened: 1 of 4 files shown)


packages/dx-grid-core/src/plugins/table-group-row/computeds.ts
```diff
@@ -114,12 +114,10 @@ export const tableGroupCellColSpanGetter: GroupCellColSpanGetter = (
 
   if (tableRow.type === TABLE_GROUP_TYPE) {
     const colName = tableColumn.column?.name;
-    const dataColumnGroupedBy =
-      tableRow.row.groupedBy === colName && tableColumn.type !== TABLE_GROUP_TYPE;
     const chains = groupSummaryChains(
       tableRow, tableColumns, groupSummaryItems, firstVisibleColumnIndex,
     );
-    const chain = chains.find(ch => !dataColumnGroupedBy && ch[0] === colName);
+    const chain = tableRow.row.groupedBy ? chains.find(ch => ch[0] === colName) : undefined;
```
torrac12, line 120: I have an question. 
when "tableRow.type === TABLE_GROUP_TYPE", if this mean "tableRow.row.groupedBy" always have an value?

I think we should only overwrite the colspan of `group type column`

torrac12: I think if there are multiple column type which have same column name,  we should get the colspan according to it's type.
Maybe we should use key of column in chains but not column name, to prevent there are mutiple chains start with same column name.

If a column is data type,  the colspan of it should not be the default value 1,  that will make the number of `td` less than we wanted,  because in some case render `null` instead `templatePlaceholder` in tableGroupRow plugin.

MikeVitik (approved):

PR #3493, opened 2022-01-28
ci: move CI to GitHub actions workflow

jacksonrr3:


.github/workflows/ci_tests.yml
```diff
@@ -14,6 +14,90 @@ jobs:
...
+
+    - name: Unpack artifacts
+      run: 7z x -y artifacts.zip
+
+    - name: TestCafe Tests Grid
+      run: yarn testcafe:ci:${{ matrix.component }}
+
+  site:
+    runs-on: ubuntu-latest
+    needs: test_cafe
```
alexserov, line 78: Looks like we depend on `build` artifacts, we can change the value of the `needs` field to speed up workflow
```suggestion
    needs: build
```

.github/workflows/ci_tests.yml
```diff
@@ -14,6 +14,90 @@ jobs:
...
+        chmod -R 777 .
+
+    - name: Linter & Jest Tests
+      run: |
+        yarn lint
+        yarn test:ci
+
+    - name: Zip artifacts 
+      run: |
+        7z a -tzip -mx3 -mmt2 artifacts.zip packages
```
alexserov, line 36: Since we execute single-line command, you can use the following:
```suggestion
      run: 7z a -tzip -mx3 -mmt2 artifacts.zip packages
```

alexserov (approved):

Krijovnick (approved):

PR #3552, opened 2022-06-06
fix(react-core): fix dragging on unmount component (T1093143)

Krijovnick:


packages/dx-react-core/src/draggable.tsx
```diff
@@ -91,20 +108,31 @@ export class Draggable extends React.PureComponent<DraggableProps> {
     e[draggingHandled] = true;
   }
```
MikeVitik, line 109: change name to isMouseEvent

packages/dx-react-core/src/draggable.tsx
```diff
@@ -91,20 +108,31 @@ export class Draggable extends React.PureComponent<DraggableProps> {
...
+      name,
+    };
+  }
+
   globalListener([name, e]) {
     switch (name) {
       case 'mousemove':
+        this.saveEvent(e, name);
         this.mouseStrategy.move(e);
         break;
```
MikeVitik, line 123: this.eventParams = null

MikeVitik (changes requested):

MikeVitik (approved):

groshenkovamarina (approved):

PR #3573, opened 2022-09-29
fix(core): fix updating plugins

### LINKED ISSUE #3570: React Chart crashes when dynamically adding Series
### Is there an existing issue for this?

- [X] I have searched this repository's [issues](https://github.com/devexpress/devextreme-reactive/issues) and believe that this is not a duplicate.

### I'm using ...

React Chart

### Current Behaviour

When creating a `<Chart/>` in React chart, I have noticed that if I try to dynamically change which series are displayed, the chart will disappear and a large error trace will appear in the console. 

It appears to only occur after the chart is created.

image

See full example in the steps to reproduce

### Expected Behaviour

I believe the correct behavior for this operation is to have the chart render the new series without error.

### Steps to Reproduce

```typescript
import {BarSeries} from "@devexpress/dx-react-chart";
import {Chart} from "@devexpress/dx-react-chart-material-ui";
import * as React from "react";
import {useEffect, useState} from "react";


interface ExampleItem {
    xAxis: number,
    series1: number,
    series2: number,
}

const chartData: ExampleItem[] = [
    {
        xAxis: 100,
        series1: 10,
        series2: 10,
    }
];

export function ExampleComp() {

    // Create a React state to keep track of which Series to show.
    const [seriesToShow, setSeriesToShow] = useState([
        "series1"
    ]);

    // Wait 2 seconds, then update the seriesToShow to have 2 series instead of 1.
    // In real world this could be something else like a button click, ect
    useEffect(() => {
        setTimeout(() => {
            setSeriesToShow([
                "series1",
                "series2"
            ])
        }, 2000)
    }, [])

    return (
        <Chart data={chartData}>

            {/* A simple loop to add all series in seriesToShow */}
            {seriesToShow.map(series => (
                <BarSeries
                    key={series}
                    argumentField="xAxis"
                    valueField={series}
                />
            ))}

        </Chart>
    );
}

const container = document.getElementById('root');
const root = createRoot(container!);
root.render(<ExampleComp />);

```

### Environment

- devextreme-reactive: 3.0.5
- react: 18.2.0
- browser:  Google Chrome (Version 105.0.5195.52 (Official Build) (arm64)), also happens in Firefox, and Safari.
- bootstrap: none
- react-bootstrap: none
- material-ui: 5.10.4

Krijovnick:
Fixes #3570

### DIFF
--- a/packages/dx-core/src/utils.ts
+++ b/packages/dx-core/src/utils.ts
@@ -17,7 +17,21 @@ export const insertPlugin = (array, newItem) => {
   const targetIndex = nextItemIndex < 0 ? array.length : nextItemIndex;
   const alreadyExists = (targetIndex >= 0 && targetIndex < array.length)
     && compare(newItem, array[targetIndex]) === 0;
-  result.splice(targetIndex, alreadyExists ? 1 : 0, newItem);
+  let deletedItems = 0;
+  if(alreadyExists) {
+    deletedItems = 1;
+    for(let i = targetIndex + 1; i < result.length; i++) {
+      const targetItemPosition = result[targetIndex].position();
+      const itemPosition = result[i].position();
+      if(targetItemPosition.length > 1 && targetItemPosition.length === itemPosition.length && 
+        targetItemPosition[targetItemPosition.length - 2] === itemPosition[itemPosition.length - 2]) {
+        deletedItems++;
+      } else {
+        break;
+      }
+    }
+  }
+  result.splice(targetIndex, deletedItems, newItem);
   return result;
 };
 
--- a/packages/dx-core/src/utils.ts
+++ b/packages/dx-core/src/utils.ts
@@ -18,14 +18,15 @@ export const insertPlugin = (array, newItem) => {
   const alreadyExists = (targetIndex >= 0 && targetIndex < array.length)
     && compare(newItem, array[targetIndex]) === 0;
   let deletedItems = 0;
-  if(alreadyExists) {
+  if (alreadyExists) {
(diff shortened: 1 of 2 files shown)


packages/dx-core/src/utils.ts
```diff
@@ -17,7 +17,22 @@ export const insertPlugin = (array, newItem) => {
   const targetIndex = nextItemIndex < 0 ? array.length : nextItemIndex;
   const alreadyExists = (targetIndex >= 0 && targetIndex < array.length)
     && compare(newItem, array[targetIndex]) === 0;
-  result.splice(targetIndex, alreadyExists ? 1 : 0, newItem);
+  let deletedItems = 0;
+  if (alreadyExists) {
+    deletedItems = 1;
+    for (let i = targetIndex + 1; i < result.length; i += 1) {
+      const targetItemPosition = result[targetIndex].position();
```
MikeVitik, line 24: Is it changed in the loop? If not move it from loop block

packages/dx-core/src/utils.ts
```diff
@@ -17,7 +17,22 @@ export const insertPlugin = (array, newItem) => {
...
+      const itemPosition = result[i].position();
+      if (targetItemPosition.length > 1 && targetItemPosition.length === itemPosition.length &&
```
MikeVitik, line 26: The same thing for
targetItemPosition.length > 1 && targetItemPosition.length === itemPosition.length

MikeVitik (changes requested):

PR #3590, opened 2022-11-01
chore: update type script

Krijovnick:


packages/dx-react-core/src/ref-holder.tsx
```diff
@@ -7,15 +7,16 @@ export const RefHolder = React.forwardRef((
...
-      ? React.cloneElement(children, { ref })
-      : React.cloneElement(children, { forwardedRef: children.props?.forwardedRef ? (node) => {
-        children.props.forwardedRef(node);
-        if (typeof ref === 'function') {
-          ref(node);
-        } else if (ref) {
-          ref.current = node;
-        }
-      } : ref }) :
+      ? React.cloneElement(children as React.ReactElement<any>, { ref })
```
MikeVitik, line 10: Is it possible to replcae
 children: React.ReactNode 
by 
children: React.ReactElement<any>

packages/dx-scheduler-core/src/constants.ts
```diff
@@ -18,21 +18,23 @@ export const POSITION_END = 'end';
 
 export const AUTO_HEIGHT = 'auto';
 
-export const DAY_OPTIONS = { day: 'numeric' };
+export const DAY_OPTIONS = { day: 'numeric' } as const;
 export const WEEK_DAY_OPTIONS = { weekday: 'short' };
```
MikeVitik, line 22: Why this const isn't as const?

MikeVitik (changes requested):

dxvladislavvolkov (approved):

MikeVitik (approved):

timbset (approved):

PR #3621, opened 2023-01-11
fix: Change <Sizer /> component width and height to 100%.

Masumi-SAJIKI:
Hello.
Thank you for providing us with a nice library.
I am using this library.

I have a problem that when I try to print an A4 size screen with a graph displayed, it is reduced to about two thirds of its original size.
Chrome version 108 or later seems to cause this problem.

A sample is below.
https://codesandbox.io/s/affectionate-hopper-rmfy8m?file=/demo.js

スクリーンショット 2023-01-11 19 47 23（2）

After investigating the cause, it appears that changing the `<Sizer />` width and height specification from 200% to 100% will eliminate the problem.

Sorry if I am rude.
Please review.

Translated with www.DeepL.com/Translator (free version)


Masumi-SAJIKI:
Sorry, I created the Issue later. #3623

Masumi-SAJIKI:
The version of Chrome has been upgraded to 110 and this issue has been resolved, so we are terminating this service.
We apologize for any inconvenience caused.

PR #3642, opened 2023-03-17
chore: use react-popper 2.x in bootstrap4 packages (permits React 18 support)

glsignal:
Hi there!

With React 18 having been available for some time now, one thing holding back an upgrade is the version of Popper in use in the bootstrap4 packages here (since popper 1.x only declares compatibility with up to React 17).

Since Popper 2.x is both compatible with React 18 and [still exposes the render-props](https://popper.js.org/react-popper/v2/render-props/) for compatibility with the existing class-based components in here, is it possible / reasonable to update the package to the 2.x line?

### DIFF
--- a/packages/dx-react-bootstrap4/components/popover.jsx
+++ b/packages/dx-react-bootstrap4/components/popover.jsx
@@ -77,7 +77,7 @@ export class Popover extends React.PureComponent {
   renderPopper() {
     const {
       children, target, renderInBody,
-      arrowComponent: ArrowComponent, modifiers,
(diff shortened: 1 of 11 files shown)


Krijovnick (approved):

ivanblinov2k17 (approved):

VasilyStrelyaev:
@glsignal, thank you for your contribution! I have added several commits to this pull request and now I'm merging it. The changes introduced in this PR will be released with the next minor update.

glsignal:
Thank you for taking the time (and extra changes) to integrate this! It's a big help 🙂

Issue #3647, opened 2023-04-12
Sorting Column Extensions not removing previous sort column


mmatuk:
### Is there an existing issue for this?

- [X] I have searched this repository's [issues](https://github.com/devexpress/devextreme-reactive/issues) and believe that this is not a duplicate.

### I'm using ...

React Grid

### Current Behaviour

Currently, when you pass `columnExtensions` to the `SortingState` plugin, you must supply `sortingEnabled` for every column that you passed. If you forget to pass `sortingEnabled: true`, even though sorting is enabled for all columns by default, sorting breaks and does not remove the sorting of the previously selected column. 

This is the column extension: 
`[
    { columnName: "name" },
    { columnName: "gender", sortingEnabled: false },
    { columnName: "city" },
    { columnName: "car" }
  ]`

This is what happens when you click on multiple columns to change sorting:
![image](https://user-images.githubusercontent.com/9585972/231536472-e8288bf6-b665-44a6-af16-d6465e2317d6.png)

This looks to be caused by a change to the base `getPersistentSortedColumns` function. Instead of looking for explicit false, it now just looks for falsely values which means `sortingEnabled == undefined` now triggers this code. 

This file `packages/dx-grid-core/src/plugins/sorting-state/helpers.ts` from this commit https://github.com/DevExpress/devextreme-reactive/commit/629e9057c0562c7d11fc5400f725555ab327fc74 looks to be the cause of the issue.

### Expected Behaviour

Prior to version 3.0.0, you could pass something like this:
`[
    { columnName: "name" },
    { columnName: "gender", sortingEnabled: false },
    { columnName: "city" },
    { columnName: "car" }
 ]`

and the sorting would handle it correctly and default to `sortingEnabled == true` for any column that did not have a `sortingEnabled` config set in the `columnExtensions`. 

I should be able to pass `columnExtensions` and not have to provide a config for every column. The function `getPersistentSortedColumns` should respect the `columnSortingEnabled` prop that is passed to the main `SortingState` plugin. Currently is does not. 

### Steps to Reproduce

See this example: https://codesandbox.io/s/grid-table-sorting-2jok85?file=/demo.js:708-848

1. Click on different column headers to change the sorting and notice it does not remove the previous sorting and that the sorting does not update correctly either.
2. You can also, change the versions of gird back to something before 3.0.0 and redo step 1 to see how it use to work. 

### Environment

- devextreme-reactive: 3.0.6
- react: 18.0
- browser: chrome
[...]

PR #3653, opened 2023-06-05
fix: regression with SortingState.columnExtensions.sortingEnabled (#3647)

VasilyStrelyaev:


Krijovnick (approved):

packages/dx-grid-core/src/plugins/sorting-state/helpers.test.ts
```diff
@@ -38,6 +38,20 @@ describe('SortingState helpers', () => {
       const result = getPersistentSortedColumns(sorting, columnExtensions);
       expect(result).toEqual(['b']);
     });
+    it('should honor the default sortingEnabled state', () => {
```
ivanblinov2k17, line 41: press F to pay respects)) i think a slight rename of test should be good

packages/dx-grid-core/src/plugins/sorting-state/helpers.ts
```diff
@@ -15,7 +15,8 @@ export const getColumnSortingDirection: GetColumnSortingDirectionFn = (
 export const getPersistentSortedColumns: GetPersistentSortedColumnsFn = (
   sorting, columnExtensions = [],
 ) => columnExtensions.reduce((acc, { columnName, sortingEnabled }) => {
-  if (!sortingEnabled) {
+  // tslint:disable-next-line:no-boolean-literal-compare
+  if (sortingEnabled === false) {
```
ivanblinov2k17, line 19: maybe we disable it not only for this line but where the strictNullChecks are false
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unnecessary-boolean-literal-compare.md#not-to-use-it

PR #3661, opened 2023-07-27
fix: conditional type exports (T1177805)

VasilyStrelyaev:
Testing with https://github.com/arethetypeswrong/arethetypeswrong.github.io, before/after:

image

image

### DIFF
--- a/scripts/merge-dts.js
+++ b/scripts/merge-dts.js
@@ -1,10 +1,10 @@
 import { join, sep } from 'path';
-import { readFileSync, existsSync } from 'fs';
+import { existsSync } from 'fs';
 import rimraf from 'rimraf';
 import replace from 'replace-in-file';
 import dts from 'dts-bundle';
 
-import { copyCommonJsTypes } from './utils.js';
+import { copyCommonJsTypes, getPackageInfo, getDtsOutFile } from './utils.js';
 
 const getIndexDts = (packageDirectory, dtsPath) => {
   let indexDts = join(dtsPath, 'index.d.ts');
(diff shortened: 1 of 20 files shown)


const314 (approved):

scripts/utils.js
```diff
@@ -0,0 +1,9 @@
+import { existsSync, copyFileSync } from 'fs';
+
+const getCommonJsTypesPath = dtsPath => dtsPath.replace(/.ts$/, '.cts');
```
timbset, line 3:
```suggestion
const getCommonJsTypesPath = dtsPath => dtsPath.replace(/\.ts$/, '.cts');
```

In your version a dot means any symbol

timbset (changes requested):

timbset (approved):

scripts/merge-dts.js
```diff
@@ -21,8 +21,8 @@ export default (packageDirectory, skipBundle = false) => {
   const dtsPath = join(packageDirectory, 'dist', 'dts');
 
   if (!skipBundle) {
-    const pkg = JSON.parse(readFileSync(join(packageDirectory, 'package.json')));
-    const dtsOutFile = join(packageDirectory, pkg.types);
+    const pkg = getPackageInfo(packageDirectory);
+    const dtsOutFile = getDtsOutFile(packageDirectory, pkg.types);
```
timbset, line 25: Looks like an overcomplicating. What's the difference between your function and using `join`?

scripts/utils.js
```diff
@@ -7,3 +8,7 @@ export const copyCommonJsTypes = (dtsPath) => {
     copyFileSync(dtsPath, getCommonJsTypesPath(dtsPath));
   }
 };
+
+export const getPackageInfo = packageDir => JSON.parse(readFileSync(join(packageDir, 'package.json')));
+
+export const getDtsOutFile = (packageDir, typesDir) => join(packageDir, typesDir);
```
timbset, line 14: We've simply duplicated using of `join`. Remove this function

scripts/merge-dts.js
```diff
@@ -21,8 +21,8 @@ export default (packageDirectory, skipBundle = false) => {
   const dtsPath = join(packageDirectory, 'dist', 'dts');
 
   if (!skipBundle) {
-    const pkg = JSON.parse(readFileSync(join(packageDirectory, 'package.json')));
-    const dtsOutFile = join(packageDirectory, pkg.types);
+    const pkg = getPackageInfo(packageDirectory);
```
timbset, line 24: I think it is better to split this on steps instead of extracting this logic into util function

timbset (changes requested):

PR #3662, opened 2023-08-04
fix(react-grid): T1176840 - Remove focus from elements in virtualRows that are not in the viewPort

ivanblinov2k17:


packages/dx-react-grid/src/components/table-layout/virtual-table-layout.tsx
```diff
@@ -271,6 +274,21 @@ export class VirtualTableLayout extends React.PureComponent<PropsType, VirtualTa
     });
   }
 
+  unfocuseVirtualRowsChildren() {
```
VasilyStrelyaev, line 277: minor, spelling: `unfocus`

packages/dx-react-grid/src/utils/helpers.ts
```diff
@@ -4,3 +4,12 @@ export const getRowStyle = ({ row }) => (row.height !== undefined
 
 export const isNumber = (value: string | number) =>
   typeof value === 'number' || !Number.isNaN(Number(value));
+
+export const recursiveBlur = (element: Element) => {
```
VasilyStrelyaev, line 8: please add tests, at least a couple of unit tests

packages/dx-react-grid/src/components/table-layout/virtual-table-layout.tsx
```diff
@@ -271,6 +274,21 @@ export class VirtualTableLayout extends React.PureComponent<PropsType, VirtualTa
...
+    const [lowerBoundary, upperBoundary] = this.props.viewport.rows;
+
+    const rowRefArray = Array.from(this.rowRefs, (entry) => {
```
VasilyStrelyaev, line 280: I usually use Object.entries/keys/values for this, assuming it should be faster to iterate maps than Array.from.
However, I may be wrong and it does not work with Map or is not actually faster...

PR #3666, opened 2023-11-06
chore: Fix packages building on Node.js v18+

timbset:


VasilyStrelyaev (approved):

VasilyStrelyaev:
* Removed the Node.js env
* Updated webpack to support new Node.js
* Added empty `act()` in chart tests to suppress errors from Popper: https://floating-ui.com/docs/react#testing

PR #3667, opened 2024-01-26
fix(grid-core): - Generate .d.ts files for grid and scheduler core packages

ivanblinov2k17:


packages/dx-scheduler-core/src/utils.ts
```diff
@@ -89,14 +89,17 @@ export const getAppointmentStyle: PureComputed<
...
-}) => ({
-  height,
-  width: `${width}%`,
-  transform: `translateY(${top}px)`,
-  msTransform: `translateY(${top}px)`,
-  left: `${left}%`,
-  position: 'absolute',
-});
+}) => {
+  const transform = `translateY(${top}px)` as any;
```
timbset, line 93: Why do you use `any` here? Could you replace it with the `string` type?

packages/dx-scheduler-core/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "outDir": "./dist",
     "downlevelIteration": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
timbset, line 6: If you succeed in avoiding `any` in a code, please, rollback the change

packages/dx-react-grid/api/dx-react-grid.api.md
```diff
@@ -1558,9 +1558,9 @@ export class TableSummaryRowBase extends React_2.PureComponent<TableSummaryRowPr
     // (undocumented)
     static GROUP_ROW_TYPE: symbol;
     // (undocumented)
-    render(): JSX.Element;
+    render(): React_2.JSX.Element;
     // (undocumented)
-    renderContent(column: any, columnSummaries: any): JSX.Element;
+    renderContent(column: any, columnSummaries: any): React_2.JSX.Element;
```
timbset, line 1563: What happened here? Why docs have changed?

timbset (changes requested):

packages/dx-scheduler-core/src/utils.ts
ivanblinov2k17: No i can't
image

packages/dx-scheduler-core/tsconfig.json
ivanblinov2k17: I haven't succeed in this. It involves hundreds of types, and i don't see the point in creating them for this task

packages/dx-react-grid/api/dx-react-grid.api.md
ivanblinov2k17: Probably newer version of api lib fixed this. In dx-react-grid.api file there are many occurencies of React_2 prefix so it shouldn't affect the documentation page

timbset (approved):

VasilyStrelyaev (approved):

PR #3677, opened 2024-08-27
fix: Reading localStorage for SSR mode

GoodDayForSurf:


VasilyStrelyaev (approved):
I still slightly prefer mocking `localStorage` for SSR like here:
https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-demo-shell/src/demo-viewer/theme-viewer.jsx#L13-L16

That's because with the current implementation we always set the `localStorage` flag to false, and then render again and reset it to `true` if the banner is displayed.

Mocking and keeping Rochmar's code should allow to avoid this

VasilyStrelyaev (approved):

PR #3691, opened 2025-04-17
fix: editing time of secondary recurring appointment

VasilyStrelyaev:


packages/dx-scheduler-core/src/plugins/editing-state/helpers.ts
```diff
@@ -137,6 +137,44 @@ export const deleteCurrentAndFollowing: DeleteFn = appointmentData => changeCurr
...
+    return moment(parentDate).add(diff).toDate();
+  };
+
+  if (changes.startDate) {
+    parentChanges = {
+      ...parentChanges,
+      startDate: convert(
+        changes.startDate as Date,
+        appointmentData.startDate as Date,
+        appointmentData.parentData.startDate as Date,
```
ivanblinov2k17, line 159: Why do we have so much typecasting as Date? Could it be moved to type def?

